PR #76, opened 2023-03-11
Enable Pytorch GPU acceleration for M1 chips + extract out Jupyter notebook setup script

### LINKED ISSUE #54: Enable Pytorch GPU acceleration for M1 chips
https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/

Docs here https://pytorch.org/docs/stable/notes/mps.html?highlight=mps

This will likely need to be done by someone with an M1.

Acceptance criteria:
- On M1 devices, Pytorch uses MPS
- Otherwise use CUDA (when available) or CPU
- Unit tests to verify

Related work: 
- Handle device types in a method we import, something like
- `from maze_transformer.experiment_imports import *
initialize_notebook(dark_mode=True, interactive_visualizations=True, rng_seed=42, on_colab=True)`

luciaquirke:
Closes #54 

- Add shared configure_notebooks method with shared configuration steps
- Enable MPS for eligible devices
- Set standard and np random seeds

There is probably more functionality that could be extracted into methods once this is merged, but for now I'm sticking to logic that can definitely be hidden without compromising notebook comprehensibility. I can't think of any meaningful unit tests for setting the device but it gets set to MPS on my machine so I think it works.

Notes:

- [x] Using a utils module over a utils class because apparently they're both conventional and a class is strictly unnecessary
- [x] Moved `torch.set_grad_enabled(False)` back to the notebooks because some notebooks will want to enable it.
- [x] What's WARNING: code hash mismatch for maze_ctor gen_dfs and how can I avoid it? - answer from Alex below
- [x] Keeping this file in the maze-transformer repo - we'll use this repo as a general source of truth for clean code, and the experiments repo for prototyping


luciaquirke:
Okay yeah I think you're right, I've moved the utils into maze-transformer

valedan:
@luciaquirke Yep all the reusable/clean etc code goes in here. I think we probably want notebooks in `maze-transformer` that run through most of the util code, at least in a basic way. 

Regarding the workflow problems you pointed out - perhaps we can modify the `experiments` repo to always pull and install the latest `maze-transformer` from the `main` branch (unless an earlier version is manually pinned, which we would do if we break `main`). What do you think?

valedan (approved):
LGTM! 👍 I like the organization here.

rusheb (approved):
LGTM

rusheb:
Couldn't find an existing ticket for code hash mismatch, so I raised one: 

https://github.com/AISC-understanding-search/maze-transformer/issues/83

rusheb:
FYI I just merged the LatticeMaze demo, so may be worth calling the new configure function from that notebook before merging this!

afspies:
Ready to merge! 
@luciaquirke very good points about the workflow for adjusting configuration over time. Indeed, I think it's nice to have a fairly comprehensive and clean baseline here, and then in experiments people can do messy stuff on top of it (and potentially extend with their own per-set-of-experiments dependencies etc.)

afspies (approved):
LGTM Pending Rusheb's suggestion above
@luciaquirke very good points about the workflow for adjusting configuration over time. Indeed, I think it's nice to have a fairly comprehensive and clean baseline here, and then in experiments people can do messy stuff on top of it (and potentially extend with their own per-set-of-experiments dependencies etc.)

luciaquirke:
> Regarding the workflow problems you pointed out - perhaps we can modify the experiments repo to always pull and install the latest maze-transformer from the main branch (unless an earlier version is manually pinned, which we would do if we break main). What do you think?

Yep this sounds exactly right

PR #78, opened 2023-03-11
Create LatticeMaze demo notebook

rusheb:
The first demo notebook, which aims to explain the maze representation.

There may now be some redundancy between this notebook and the docstrings I had previously written for some of these functions. Should we keep the docstrings? Should I delete them?

### DIFF
--- a/maze_transformer/generation/latticemaze.py
+++ b/maze_transformer/generation/latticemaze.py
@@ -151,7 +151,7 @@ class LatticeMaze:
 
         return img
 
-    def as_adjlist(
+    def as_adj_list(
         self, shuffle_d0: bool = True, shuffle_d1: bool = True
     ) -> NDArray["conn start_end coord", np.int8]:
         adjlist: NDArray["conn start_end coord", np.int8] = np.full(
--- a/maze_transformer/training/tokenizer.py
+++ b/maze_transformer/training/tokenizer.py
@@ -39,7 +39,7 @@ class MazeTokenizer:
                         node_token_map[tuple(c_e.tolist())],
                         SPECIAL_TOKENS["adjacency_endline"],
                     ]
-                    for c_s, c_e in self.maze.as_adjlist()
+                    for c_s, c_e in self.maze.as_adj_list()
                 ]
             ),
             SPECIAL_TOKENS["adjlist_end"],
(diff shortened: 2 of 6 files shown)


notebooks/demo_001_latticemaze.ipynb
```diff
@@ -0,0 +1,362 @@
...
+    "N   N - N - N\n",
+    "|           |\n",
+    "N - N - N - N\n",
+    "            |\n",
+    "N - N   N - N\n",
+    "|       | \n",
+    "N - N - N - N\n",
+    "```\n",
+    "\n",
+    "Each square in the visualisation, may represent a border, a node, a connection, or a wall. In each direction, there are always \n",
```
valedan, line 271: I find it a bit confusing to look at the image and have connections represented as white squares that look identical to nodes. It seems like it would make more sense to not render connections. I assume the reason we're doing that is because the walls (non-connections) take up a full square. I wonder if an alternate version of `as_img()` with thin walls and no connection squares would be clearer... 🤔

valedan (approved):
@rusheb This is great! Very helpful. Looking forward to more!

rusheb:
Wow, that's a lot of typos. Sorry/thanks everyone!

notebooks/demo_001_latticemaze.ipynb
```diff
@@ -0,0 +1,362 @@
...
+    {
+     "data": {
+      "text/plain": [
+       "LatticeMaze(lattice_dim=2, connection_list=array([[[ True, False, False,  True],\n",
+       "        [False, False, False,  True],\n",
+       "        [ True, False,  True, False],\n",
+       "        [False, False, False, False]],\n",
+       "\n",
+       "       [[False,  True,  True, False],\n",
+       "        [ True,  True,  True, False],\n",
```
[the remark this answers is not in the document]
rusheb: hmm...I think that's in the outputs. Can I do anything about it?

notebooks/demo_001_latticemaze.ipynb
rusheb: I also found this confusing. New issue created: https://github.com/AISC-understanding-search/maze-transformer/issues/84

PR #79, opened 2023-03-12
Add origin coordinate to maze tokenization

luciaquirke:
- Add missing start_pos to tokenized maze
- in SPECIAL_TOKENS most tokens come as x_start or x_end, but a few are start_x and end_x. I've updated the keys but I'm hesitant to change the values because seems like it would probably break people's datasets

Future work, not this PR:
- cfg and config used to refer to different objects, they both need descriptive names. Just need to update this in a way that doesn't break any important existing datasets / models
- add integration test that calls everything twice when and if the set_reproducibility method works, to ensure it stays working


tests/unit/maze_transformer/generation/test_generators.py
```diff
@@ -16,3 +18,35 @@ def test_gen_dfs_oblong():
     maze = LatticeMazeGenerators.gen_dfs(three_by_four)
 
     assert maze.connection_list.shape == (2, 3, 4)
+
+
+def test_gen_dfs_deterministic_with_seed():
```
luciaquirke, line 23: I might delete these, it just took me a minute to remember how seeds work and I wrote this in the meantim

valedan:
This looks good so far!

maze_transformer/generation/latticemaze.py
```diff
@@ -14,8 +14,10 @@
     adjlist_end="<ADJLIST_END>",
     target_start="<TARGET_START>",
     target_end="<TARGET_END>",
-    start_path="<START_PATH>",
-    end_path="<END_PATH>",
+    origin_start="<ORIGIN_START>",
+    origin_end="<ORIGIN_END>",
+    path_start="<START_PATH>",
+    path_end="<END_PATH>",
```
rusheb, line 20: agh! thank you! this had been bugging me as well.

Do you think it's worth changing the string as well so that start/end is the suffix?

rusheb (approved):

maze_transformer/generation/latticemaze.py
luciaquirke: Yeah if people are happy to have their trained models not be compatible with future datasets

rusheb: Worth asking on slack?

luciaquirke: Actually I just realised that adding the origin isn't backwards compatible either so I'll just go for it and announce it

rusheb (approved):
🚀

Issue #73, opened 2023-03-10
Use consistent coordinate system


rusheb:
I've been getting really confused by the different coordinate systems used for different maze representations

* The LatticeMaze connection_list, adjlist and as_img use (x is rows) with origin at top left
* the path produced by plot_path uses (x is columns)  with origin at top left (meaning we have to transpose the image to plot them together
* the path produced by plot_multi_paths uses (x is columns) with origin at bottom left!! 

We should pick a convention and stick to it. to minimise the amount of changes we need to make, I think we should use
* Origin at top left
* x is rows is down
* y is columns is right

PR #80, opened 2023-03-12
Test for adjacency list coordinates

valedan:
Based on [this comment](https://github.com/AISC-understanding-search/maze-transformer/issues/73#issuecomment-1464923828) - my understanding is that we want to test that the `LatticeMaze` and `MazeTokenizer` adjacency lists are using the same coordinate system.

@rusheb let me know if this is what you were thinking or if you had something else in mind!

It's a fairly complicated (and brittle) test so I welcome any suggestions for how this could be simplified.

Also welcoming any style nitpicks.

### DIFF
--- a/tests/helpers/utils.py
+++ b/tests/helpers/utils.py
@@ -1,0 +1,7 @@
+# We don't care about order of coordinate pairs within
+# the adjlist or coordinates within each coordinate pair.
+def adjlist_to_nested_set(adjlist):
+    return {
+        frozenset([tuple(start_coord), tuple(end_coord)])
+        for start_coord, end_coord in adjlist
+    }
--- a/tests/helpers/utils.py
+++ b/tests/helpers/utils.py
@@ -1,3 +1,5 @@
+# Used for comparison of adjlists
+# Adjlist looks like [[[0, 1], [1, 1]], [[0, 0], [0, 1]], ...]
 # We don't care about order of coordinate pairs within
 # the adjlist or coordinates within each coordinate pair.
 def adjlist_to_nested_set(adjlist):
--- a/tests/unit/maze_transformer/training/test_tokenizer.py
+++ b/tests/unit/maze_transformer/training/test_tokenizer.py
@@ -1,5 +1,3 @@
-import re
-
 import numpy as np
 
 from maze_transformer.generation.generators import LatticeMazeGenerators
@@ -40,8 +38,9 @@ def test_coordinate_system():
 
     # remove special tokens
     tokenizer_adjlist_coordinates = [
-        token for token in tokenizer_adjlist if re.search(r"\(\d+,\d+\)", token)
+        token for token in tokenizer_adjlist if token not in SPECIAL_TOKENS.values()
     ]
+
     # Group pairs of coordinates
     tokenizer_adjlist_connections = [
         *zip(tokenizer_adjlist_coordinates[::2], tokenizer_adjlist_coordinates[1::2])
--- a/tests/unit/maze_transformer/training/test_tokenizer.py
+++ b/tests/unit/maze_transformer/training/test_tokenizer.py
@@ -1,8 +1,11 @@
+import re
+
 import numpy as np
 
 from maze_transformer.generation.generators import LatticeMazeGenerators
 from maze_transformer.training.mazedataset import MazeDatasetConfig
 from maze_transformer.training.tokenizer import SPECIAL_TOKENS, MazeTokenizer
+from tests.helpers import utils
 
 
 def test_coordinate_system():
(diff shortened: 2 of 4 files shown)


tests/unit/maze_transformer/training/test_tokenizer.py
```diff
@@ -0,0 +1,55 @@
...
+    ).node_token_map
+
+    tokenized_maze = MazeTokenizer(
+        maze=maze,
+        solution=np.array(
+            maze.find_shortest_path(
+                c_start=(0, 0),
+                c_end=(maze_size - 1, maze_size - 1),
+            )
+        ),
```
rusheb, line 34: I don't think we need to compute the solution, we could just set the value to `np.array([[0,0]])`. Though maybe it's just cleaner to do it it as you have. Not sure! Thoughts?

tests/unit/maze_transformer/training/test_tokenizer.py
```diff
@@ -0,0 +1,55 @@
...
+        for i in range(0, len(tokenizer_adjlist), 4)
+    ]
+
+    # sort both adjlists
+    maze_adjlist_connections = sorted(
+        [sorted(conn) for conn in maze_adjlist_connections]
+    )
+    tokenizer_adjlist_connections = sorted(
+        [sorted(conn) for conn in tokenizer_adjlist_connections]
+    )
```
rusheb, line 53: i think this will work without the square brackets (see https://peps.python.org/pep-0289/)

```python
    maze_adjlist_connections = sorted(
        sorted(conn) for conn in maze_adjlist_connections
    )
    tokenizer_adjlist_connections = sorted(
        sorted(conn) for conn in tokenizer_adjlist_connections
    )
```

rusheb: Also, we could use `frozenset` here. Check out the tests in `test_latticemaze.py`. Might be good to use the same comparison method for consistency, but I don't know if it's important. Feel free to ignore!

rusheb (approved):
As discussed offline, this wasn't exactly the test I had in mind. Nevertheless I think it might be useful. I agree it seems possibly a bit brittle but not sure what to do about that. 

Perhaps the solution is to parse into an intermediate representation, but that seems like overkill.

Left a few nits - feel free to ignore them

tests/unit/maze_transformer/training/test_tokenizer.py
valedan: Yep good call. I'm using your function for this now.

tests/unit/maze_transformer/training/test_tokenizer.py
```diff
@@ -0,0 +1,55 @@
...
+
+    tokenizer_adjlist = tokenized_maze[
+        1 : tokenized_maze.index(SPECIAL_TOKENS["adjlist_end"])
+    ]
+
+    # There are 4 tokens per connection, only the first and third are coords
+    tokenizer_adjlist_connections = [
+        [tokenizer_adjlist[i], tokenizer_adjlist[i + 2]]
+        for i in range(0, len(tokenizer_adjlist), 4)
+    ]
```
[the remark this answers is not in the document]
valedan: Thanks, I didn't know that about range! Reworked this - it's also less brittle now because it will handle any number of special tokens per connection.

valedan:
@rusheb I addressed your feedback, and moved the `_to_nested_set` method into a helper module so I could use it too. Let me know what you think!

tests/unit/maze_transformer/training/test_tokenizer.py
```diff
@@ -0,0 +1,52 @@
...
+    ).as_tokens(node_token_map)
+
+    tokenizer_adjlist = tokenized_maze[
+        tokenized_maze.index(SPECIAL_TOKENS["adjlist_start"])
+        + 1 : tokenized_maze.index(SPECIAL_TOKENS["adjlist_end"])
+    ]
+
+    # remove special tokens
+    tokenizer_adjlist_coordinates = [
+        token for token in tokenizer_adjlist if re.search(r"\(\d+,\d+\)", token)
```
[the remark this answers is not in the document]
rusheb: Just messing around:

```python
    
    ...

    tokenizer_adjlist = tokenized_maze[
        tokenized_maze.index(SPECIAL_TOKENS["adjlist_start"])
        + 1 : tokenized_maze.index(SPECIAL_TOKENS["adjlist_end"])
    ]

    tokenizer_adjlist_connections = pairs(
        i for i in tokenizer_adjlist if i not in SPECIAL_TOKENS.values()
    )

    assert utils.adjlist_to_nested_set(
        tokenizer_adjlist_connections
    ) == utils.adjlist_to_nested_set(maze_adjlist_connections)

def pairs(t):
    it = iter(t)
    return list(zip(it, it))
```

had a bit of help from https://stackoverflow.com/questions/4628290/pairs-from-single-list

tests/unit/maze_transformer/training/test_tokenizer.py
```diff
@@ -0,0 +1,52 @@
...
+    tokenized_maze = MazeTokenizer(
+        maze=maze,
+        solution=np.array([[0, 0]]),
+    ).as_tokens(node_token_map)
+
+    tokenizer_adjlist = tokenized_maze[
+        tokenized_maze.index(SPECIAL_TOKENS["adjlist_start"])
+        + 1 : tokenized_maze.index(SPECIAL_TOKENS["adjlist_end"])
+    ]
+
```
rusheb, line 40: I realised that we probably want methods on the tokenized_maze to do this.

e.g. here we could call `tokenized_maze.adjlist()` to get just the tokens for the adjlist or `tokenized_maze.solution()` if we wanted to get just the tokens for the solution.

Not sure how well this would generalise to different tokenization schemes. Maybe it would be better to implement it using index notation e.g. `tokenized_maze["adjlist"]`, or otherwise we could have subclasses of MazeTokenizer for each tokenization scheme. Maybe there's an alternative way to do it with composition rather than inheritance?

Probably out of scope but maybe worth raising an issue if you like the idea

valedan: I do like this idea! I'll create a ticket for this, it is a bit out of scope.

tests/unit/maze_transformer/training/test_tokenizer.py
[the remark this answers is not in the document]
valedan: Great call on `if token not in SPECIAL_TOKENS`! I'll do that.

I also saw that SO thread, but decided against the `iter` approach because I found it less intuitive than just using a `zip`. Might just be my lack of experience with Python though.

tests/helpers/utils.py
```diff
@@ -0,0 +1,7 @@
+# We don't care about order of coordinate pairs within
+# the adjlist or coordinates within each coordinate pair.
```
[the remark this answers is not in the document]
valedan: Added some more context

PR #81, opened 2023-03-13
Create `HuggingMazeTokenizer` which works with `HookedTransformer`

afspies:
The tokenizer wraps the `PreTrainedTokenizer` (slow) class from HuggingFace. The tokenizer can be included in the initialization of the `HookedTransformer` allowing `model.generate(...)`, `model.to_tokens(...)`, etc. to be used. 

Inputs for tokenization should be either a string `<ADJ_START> .... (0,0) ...'` or list of such strings (as opposed to a list of strings for a single example, e.g. not `['<ADJ_START>', ... ,'(0,0)']`. Padding of batches etc. is handled by the underlying tokenizers, and numerous tests have been added to check this.

Additionally added a small `tokenizer.to_ascii()` method which returns a string representation of the maze in 2D.


tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,276 @@
...
+        ";",
+        "(0,2)",
+        "<-->",
+        "(0,1)",
+        ";",
+        "<ADJLIST_END>",
+        "<TARGET_START>",
+        "(2,1)",
+        "<TARGET_END>",
+        "<START_PATH>",
```
luciaquirke, line 85: Overlap with [my PR](https://github.com/AISC-understanding-search/maze-transformer/pull/79/files#diff-23545957e6414d237a17e96a79f5d6a5db64b00d59859a50ce776e0a202ffa14), needs the new origin tokens

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,276 @@
...
+    ), "ASCII encoding from string tokens failed"
+
+    # And with token ids
+    token_ids = tokenizer.encode(maze_str_tokens)
+    assert (
+        tokenizer.to_ascii(token_ids).splitlines() == target
+    ), "ASCII encoding from token ids failed"
+
+
+def test_inside_hooked_transformer():
```
luciaquirke, line 121: Test a single scenario in each unit test:
- allows for more parallelisation = tests run faster
- tests can have more specific names so it's faster to understand what went wrong when they fail (those comments with the -- -- formatting could be test names)
- if single purpose tests are hard to write it's a useful signal that there's something wrong with the code
- lower risk of state-related bugs from other test scenarios, e.g. if you set a variable correctly in the first scenario, then a bug in the second scenario means it doesn't get set when it's supposed to, the test might not catch it

maze_transformer/training/config.py
```diff
@@ -89,6 +90,13 @@ class TrainConfig:
         d_head=16,
         n_layers=4,
     ),
+    BaseGPTConfig(
```
luciaquirke, line 93: I would add a comment explaining how d_model, d_head, n_layers are calculated and possibly remove the math and add the final number

maze_transformer/training/tokenizer.py
```diff
@@ -60,3 +71,92 @@ def as_tokens(
             )
 
         return tokens
+
+
+class HuggingMazeTokenizer(PreTrainedTokenizer):
```
luciaquirke, line 76: nit: HuggingMazeTokenizer -> HuggingFaceMazeTokenizer

luciaquirke, line 76: Needs type hints

maze_transformer/training/tokenizer.py
```diff
@@ -60,3 +71,92 @@ def as_tokens(
...
+    vocab: dict[str, int]  # map of token_ids to strings
+
+    bos_token: str = SPECIAL_TOKENS["start_path"]
```
luciaquirke, line 79: I renamed these to `path_start` and `path_end` in the other PR to be consistent with the other special tokens, plz forgive me

maze_transformer/training/tokenizer.py
```diff
@@ -60,3 +71,92 @@ def as_tokens(
...
+    eos_token: str = SPECIAL_TOKENS["end_path"]
+    pad_token: str = SPECIAL_TOKENS["padding"]
+    unk_token: str = "<UNK>"
+
+    vocab_size: int = 0
```
luciaquirke, line 84: is this supposed to always be 0?

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,276 @@
...
+
+    # WrappedTokenizer
+    # Initialized with a configholder - tokenizer will eventually be a string
+    cfg_holder = ConfigHolder(
+        train_cfg=None, dataset_cfg=cfg, model_cfg=None, tokenizer=None
+    )
+    tokenizer = HuggingMazeTokenizer(cfg_holder)
+
+    tokenizer_out = tokenizer(maze_str_tokens)["input_ids"]
+    assert torch.all(
```
luciaquirke, line 40: Suggest `assert list(itertools.chain.from_iterable(tokenizer_out)) == maze_tokens` to avoid the torch thing

luciaquirke (commented):
This looks super helpful 🚀  I'm still building context on the high level goals here so I'll leave rusheb to handle approval or come back to this later

maze_transformer/training/tokenizer.py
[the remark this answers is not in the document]
rusheb: I'm torn on this one, because I also find them unclear, but they are consistent with [HuggingFace](https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PreTrainedTokenizer)

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,276 @@
...
+def test_inside_hooked_transformer():
+    # Check that the wrapped tokenizer facilitates all HookedTransformer tokenizer-dependent functionality
+    maze = generate_MazeTokenizer(None, 3, (0, 0), (2, 1))
+
+    # Need to generate a config to extract the token map >.<
+    cfg = MazeDatasetConfig(name="testing_maze", grid_n=3, n_mazes=1)
+    node_token_map = cfg.node_token_map
+
+    # Adjacency List Tokenization
+    maze_str_tokens = [
```
rusheb, line 130: may be more readable

```python
    maze_str_tokens = """<ADJLIST_START> (1,1) <--> (2,1) ; (2,0) <--> (1,0) ;
    (0,1) <--> (0,0) ; (2,2) <--> (2,1) ; (2,0) <--> (2,1) ; (0,2) <--> (1,2) ;
    (0,0) <--> (1,0) ; (0,2) <--> (0,1) ; <ADJLIST_END> <TARGET_START> (2,1)
    <TARGET_END> <START_PATH> (0,0) (1,0) (2,0) (2,1) <END_PATH>""".split()
```

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,276 @@
...
+    )
+    tokenizer = HuggingMazeTokenizer(cfg_holder)
+
+    tokenizer_out = tokenizer(maze_str_tokens)["input_ids"]
+    assert torch.all(
+        torch.tensor(tokenizer_out).flatten() == torch.tensor(maze_tokens)
+    ), "Tokenization mismatch"
+
+
+def test_ascii_encoding():
```
rusheb, line 45: Maybe rename `test_to_ascii` for searchability?

maze_transformer/generation/latticemaze.py
```diff
@@ -151,6 +151,33 @@ def as_img(self) -> NDArray["x y", bool]:
 
         return img
 
+    def as_ascii(self, start=None, end=None):
```
rusheb, line 154: Regarding the testing of this function - it is indirectly tested by the `test_ascii_encoding`, but that test also tests other logic. Would we want an additional test for this function? If we added one, would we want to mock out this function in the other test? 

I'm never sure about this stuff. @luciaquirke  @valedan do either of you want to chime in?

maze_transformer/training/tokenizer.py
```diff
@@ -60,3 +71,92 @@ def as_tokens(
...
+        x for x in SPECIAL_TOKENS.values() if x not in [SPECIAL_TOKENS["padding"]]
+    ]
+
+    # Overwrite class attributes
+    padding_side = "left"
+    truncation_side = "left"  #! strange choice, but it's what we did in pad_sequence
+
+    name_or_path = "maze_tokenizer"
+
+    def __init__(self, cfg: Union["ConfigHolder", "MazeDatasetConfig"], **kwargs):
```
rusheb, line 95: We are doing a lot of work in this constructor. My instinct says that we should simplify the constructor to just call super and do any assignment, and any additional work to a factory method. Maybe my mindset is too OOP. I'm not sure. @luciaquirke @valedan interested in your thoughts here too.

tests/unit/maze_transformer/generation/test_tokenizers.py
rusheb: ditto for the other token lists in these tests

rusheb (changes requested):
This looks really good! - Looks really good. My main high-level observation is that we are still having to do a lot of messy stuff with configs and serialization. Do we want to sort any of this out, or are we still waiting to move to zanj/wandb? Specific items:
- Tokenizer is not serialized on the ConfigHolder
- ConfigHolder modifies its own state (self.tokenizer property) when calling `create_model`
- Generating a MazeDatasetConfig just to extract `node_token_map` (issue 77)
- It seems fishy to me that we construct a ConfigHolder just to create the MazeTokenizer....but I remember what you said before about just wanting a single config object to pass around

once the other comments have been addressed it might be more apparent how we can tidy things up further -- for example extracting out duplicate logic in tests. 

I also haven't looked at the notebooks yet.

afspies:
I also find that super gross, and in fact think that we should probably allow the tokenizer to take a maze object directly too ; however, i figured I'd wait and see what we do with configuration and new tokenization schemes first.

rusheb:
> I also find that super gross, and in fact think that we should probably allow the tokenizer to take a maze object directly too ; however, i figured I'd wait and see what we do with configuration and new tokenization schemes first.

tokenizer taking a maze sounds like a really nice suggestion. then we can encapsulate the dataset-config-creation-just-to-get-the-node-token-map right?! that sounds worth doing!

maze_transformer/training/tokenizer.py
afspies: Yes - when we use the `PreTrainedTokenizer`'s `add_token...` methods, those "added tokens" are considered an additional part of the "base vocab" which would typically be loaded from a file. We circumvent the need to have a file containing our tokens by having an empty base vocab, such that everything is loaded during initialization from the maze config

afspies:
The only outstanding points regard breaking up the tokenizer tests and not using torch operations for equality checks. I prefer to use array checks (especially for the batched sequence), and am not sure how to let multiple tests use the same hookedtransformer and config (initializing those is the slowest part of the test by far, so it would probably be a slow down to copy paste them into multiple tests)

rusheb (approved):
LGTM

tests/unit/maze_transformer/generation/test_latticemaze.py
```diff
@@ -40,7 +40,7 @@ def test_as_img():
     assert_array_equal(expected, img)
 
 
-def test_as_adj_list():
+def test_as_adjlist():
```
luciaquirke, line 43: Surely adjacency list is two words so needs the underscore?

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,276 @@
...
+from scripts.create_dataset import generate_MazeTokenizer
+
+
+def test_tokenization_encoding():
+    # Check that wrapped tokenizer __call__ returns the same as original tokenizer
+    maze = generate_MazeTokenizer(None, 3, (0, 0), (2, 1))
+
+    # Need to generate a config to extract the token map >.<
+    cfg = MazeDatasetConfig(name="testing_maze", grid_n=3, n_mazes=1)
+    node_token_map = cfg.node_token_map
```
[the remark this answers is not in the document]
luciaquirke: This would look like `node_token_map = MazeDatasetConfig(name="testing_maze", grid_n=3, n_mazes=1).node_token_map`

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,170 @@
...
+    # Need to generate a config to extract the token map >.<
+    # TODO: Part of https://github.com/AISC-understanding-search/maze-transformer/issues/77
+    cfg = MazeDatasetConfig(name="testing_maze", grid_n=3, n_mazes=1)
+    node_token_map = cfg.node_token_map
+
+    # Adjacency List Tokenization
+    maze_str_tokens = maze.as_tokens(node_token_map)
+
+    # Manual Tokenization
+    vocab_map = {token: i for i, token in enumerate(cfg.token_arr)}
```
luciaquirke, line 30: Sometimes a variable name with a type in it (Hungarian notation) can be renamed to be even more informative, e.g. instead of vocab_map it could be vocab_to_index, which gives you information about the value as well as the key

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,170 @@
...
+    maze_tokens = [vocab_map[token] for token in maze_str_tokens]
```
luciaquirke, line 31: maze_vocab_indices

tests/unit/maze_transformer/generation/test_tokenizers.py
```diff
@@ -1,5 +1,170 @@
...
+    ), "ASCII encoding from string tokens failed"
+
+    # And with token ids
+    token_ids = tokenizer.encode(maze_str_tokens)
+    assert (
+        tokenizer.to_ascii(token_ids).splitlines() == target
+    ), "ASCII encoding from token ids failed"
+
+
+def test_inside_hooked_transformer():
```
luciaquirke, line 81: test_inside_hooked_transformer -> test_tokenizer_inside_hooked_transformer

luciaquirke (approved):
Looking good!

luciaquirke:
@afspies I haven't tried it but maybe we could define a fixture for each of hookedtransformer and config in `conftest.py` like

```
@pytest.fixture()
def hooked_transformer() -> HookedTransformer:
    return HookedTransformer(whatever)
```

then use like
```
@pytest.mark.usefixtures("hooked_transformer")
test_x_y_z(hooked_transformer):
```

maze_transformer/training/tokenizer.py
```diff
@@ -64,3 +74,94 @@ def as_tokens(
...
+
+    def __call__(self, text, **kwargs) -> BatchEncoding:
+        """
+        Tokenizer will take a list of strings and encode each
+        I.e. a single example should be a continuous string
+            "a b c d e f" not ["a", "b", "c", "d", "e", "f"]
+        """
+        try:
+            if isinstance(text, LatticeMaze):
+                text = text.as_tokens(node_token_map=self.vocab)
```
valedan, line 130: `LatticeMaze` doesn't have an `as_tokens` method. Did you mean to use the `MazeTokenizer` here?

afspies:
Any last comments

luciaquirke:
adjlist -> adj_list

image

rusheb:
wow @luciaquirke I can't believe you used 1 of your 25 messages to GPT-4 on that

luciaquirke:
I just like to flex it sometimes 😊

luciaquirke:
Nah I never fixed it. In general it seems fine to update the tokenization scheme as long as we post in the chat beforehand so people aren't caught unawares - we can continue to explore models trained on old tokenization schemes by using old datasets.

PR #82, opened 2023-03-13
Add test for create dataset script

luciaquirke:
- Add a few basic tests for create dataset
- Add comment for a class that I didn't understand at first

### DIFF
--- a/scripts/create_dataset.py
+++ b/scripts/create_dataset.py
@@ -18,6 +18,14 @@ def generate_MazeTokenizer(
     c_start: tuple[int, int],
     c_end: tuple[int, int],
 ) -> MazeTokenizer:
+    """
+    Generates and solves a maze then wraps the maze and its solution in a MazeTokenizer for serialization
+
+    Parameters:
+        grid_n (int): defines both the height and the width of the maze
+        c_start (tuple[int, int]): starting coordinate of the maze
+        c_end (tuple[int, int]): ending coordinate of the maze
+    """
     maze = LatticeMazeGenerators.gen_dfs(
         grid_shape=(grid_n, grid_n),
         lattice_dim=2,
@@ -41,7 +49,9 @@ def create_dataset(
     **cfg_kwargs,
 ):
     if n_mazes < 0:
-        raise ValueError("m_mazes must be >= 0")
+        raise ValueError("n_mazes must be >= 0")
+    if grid_n < 0:
+        raise ValueError("grid_n must be >= 0")
 
     name_base: str = (
         f"g{grid_n}-n{shorten_numerical_to_str(n_mazes, small_as_decimal = False)}"
--- a/tests/integration/scripts/test_create_dataset.py
+++ b/tests/integration/scripts/test_create_dataset.py
@@ -1,0 +1,49 @@
+import os
+import tempfile
+from pathlib import Path
+
+import pytest
+
+from maze_transformer.training.mazedataset import MazeDataset
+from scripts.create_dataset import create_dataset
+
+
+@pytest.fixture
+def temp_dir() -> Path:
+    data_dir = tempfile.TemporaryDirectory()
+    yield Path(data_dir.name)
+    data_dir.cleanup()
+
+
+def test_create_expected_files_and_directories(temp_dir):
+    n_mazes = 5
+    grid_n = 3
+    name = "test"
+
+    create_dataset(path_base=str(temp_dir), n_mazes=n_mazes, grid_n=grid_n, name=name)
+    dataset_directory_name = f"g{grid_n}-n{n_mazes}-{name}"
+    files = MazeDataset.DISK_SAVE_FILES
+    file_names = {
+        value for attr, value in vars(files).items() if not attr.startswith("__")
+    }
+
+    assert os.path.isdir(os.path.join(temp_dir, dataset_directory_name))
+    for file_name in file_names:
+        assert os.path.isfile(os.path.join(temp_dir, dataset_directory_name, file_name))
+
+
+def test_invalid_n_mazes_values(temp_dir):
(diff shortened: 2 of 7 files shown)


scripts/create_dataset.py
```diff
@@ -42,6 +50,8 @@ def create_dataset(
 ):
     if n_mazes < 0:
         raise ValueError("m_mazes must be >= 0")
```
valedan, line 52: `m_mazes` -> `n_mazes`

scripts/create_dataset.py
```diff
@@ -42,6 +50,8 @@ def create_dataset(
...
+    if grid_n < 0:
+        raise ValueError("m_mazes must be >= 0")
```
valedan, line 54: `m_mazes` -> `grid_n`

tests/unit/scripts/test_create_dataset.py
```diff
@@ -0,0 +1,41 @@
+import os
```
valedan, line 1: These tests look good. 👍

tests/unit/scripts/test_create_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+def temp_dir() -> Path:
+    data_dir = tempfile.TemporaryDirectory()
+    yield Path(data_dir.name)
+    data_dir.cleanup()
+
+
+def test_create_expected_files_and_directories(temp_dir):
+    create_dataset(path_base=str(temp_dir), n_mazes=5, grid_n=3, name="test")
+    dataset_directory_name = "g3-n5-test"
+    file_names = {"cfg.json", "maze_tokenized.npz", "maze_tokens.jsonl"}
```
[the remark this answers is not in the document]
valedan: Yeah I think something along those lines would be nice, to avoid duplicating these filenames. But I'm not sure the best way to make it tidy.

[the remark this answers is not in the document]
rusheb: something like this? a bit gross, but does the job
```python
files = MazeDataset.DISK_SAVE_FILES
file_names = {value for attr, value in vars(d).items() if not attr.startswith("__")}
```

rusheb (commented):
Thanks for adding these! 

If we are adding unit tests tests for `create_dataset`, that makes me think maybe we should move it out of `scripts`. Thoughts?

tests/unit/scripts/test_create_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+import tempfile
+from pathlib import Path
+
+import pytest
+
+from scripts.create_dataset import create_dataset
+
+
+@pytest.fixture
+def temp_dir() -> Path:
```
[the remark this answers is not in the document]
luciaquirke: Yep let's move them all to integration, very keen to keep the unit tests speedy

tests/unit/scripts/test_create_dataset.py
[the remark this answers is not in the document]
luciaquirke: Oh that looks great, thanks :)

luciaquirke:
@rusheb I'm not totally sure, in general I think it's fine to have tests for scripts but this does seem like core functionality. If you have a strong opinion on what exactly we should do I'll follow your lead, otherwise I'll add the tests now then wait until I have more context on the codebase to form an opinion. I can see the method evolving a fair amount as we add features like varying the origin and target coordinates in the meantime.

My current low-confidence thinking is that this code seems similar to the MazeDataset.gen_default method but with different input parameters. Maybe we could extract everything except the IO and CLI parts into a core method that returns a MazeDataset, then try to use that method in both MazeDataset.gen_default  and create_dataset?

tests/integration/scripts/test_create_dataset.py
```diff
@@ -0,0 +1,49 @@
...
+    n_mazes = 5
+    grid_n = 3
+    name = "test"
+
+    create_dataset(path_base=str(temp_dir), n_mazes=n_mazes, grid_n=grid_n, name=name)
+    dataset_directory_name = f"g{grid_n}-n{n_mazes}-{name}"
+    files = MazeDataset.DISK_SAVE_FILES
+    file_names = {
+        value for attr, value in vars(files).items() if not attr.startswith("__")
+    }
```
rusheb, line 28: super unimportant, but I actually think it would be nicer to inline the variable

rusheb (approved):
🚀

rusheb:
> @rusheb I'm not totally sure, in general I think it's fine to have tests for scripts but this does seem like core functionality. If you have a strong opinion on what exactly we should do I'll follow your lead, otherwise I'll add the tests now then wait until I have more context on the codebase to form an opinion on moving this out of `scripts`. I can see the method evolving a fair amount in the meantime as we add features like varying the origin and target coordinates.
> 
> My current low-confidence thinking is that this code seems similar to the MazeDataset.gen_default method but with different input parameters. Maybe we could extract everything except the IO and CLI parts into a core method that returns a MazeDataset, then try to use that method in both MazeDataset.gen_default and create_dataset?

I don't have a strong opinion either tbh. I like the idea of keeping scripts to IO and CLI where possible. 

Good spot with `gen_default` - I don't think I've even looked that method before. As far as I can tell it isn't used anywhere? So maybe I would go as far as to say to replace gen_default with the logic from `create_dataset`. We might want to rename the method too.

I'm happy to come back to it. Optionally we could create a ticket to track.

tests/integration/scripts/test_create_dataset.py
luciaquirke: It's the autoformatter 😂

luciaquirke:
Tracked here: https://github.com/orgs/AISC-understanding-search/projects/1/views/1?pane=issue&itemId=22957839

PR #85, opened 2023-03-13
73 - Fix coordinates in plot_maze

### LINKED ISSUE #73: Use consistent coordinate system

valedan:
I believe this resolves #73 

The issue was that in our generation and tokenization code we are using array notation for all coordinates - (row, col). But matplotlib uses Cartesian coordinates - (x, y). These are opposites, which caused a lot of confusion in `plot_maze.py` where we are plotting mazes and paths.

There are 2 fixes here:
- Swap the dimensions of the path coordinates when we plot them in matplotlib
- Fix usage of `extent` which was causing origin to appear at the bottom instead of the top.

**Note**: I don't have any tests for this yet. Both of these functions are just rendering plots so unit tests are a bit of a pain. I'll have to extract out all the logic involved in calculating the inputs for `plt.plot`  into seperate functions and test those. Putting the PR up now so I can get a sanity check on the actual fix while I work on tests and refactoring.

Plots after fix:

Screenshot 2023-03-13 at 3 37 52 PM
Screenshot 2023-03-13 at 4 01 03 PM

### DIFF
--- a/maze_transformer/evaluation/plot_maze.py
+++ b/maze_transformer/evaluation/plot_maze.py
@@ -44,15 +44,14 @@ def plot_multi_paths(
     show: bool = True,
 ) -> None:
     # show the maze
-    img = maze.as_img()
     plt.imshow(
-        np.rot90(img, 1),
+        maze.as_img(),
         cmap="gray",
         extent=[
             -0.75,
             maze.grid_shape[0] - 1 + 0.75,
-            -0.75,
             maze.grid_shape[1] - 1 + 0.75,
+            -0.75,
         ],
     )
 
@@ -63,10 +62,12 @@ def plot_multi_paths(
 
         # p_transformed: NDArray = maze.points_transform_to_img(pf.path)
         p_transformed: NDArray = np.array(pf.path)
-
         if pf.quiver_kwargs is not None:
-            x: NDArray = p_transformed[:, 0]
-            y: NDArray = p_transformed[:, 1]
+            # Pyplot uses Cartesian coordinates (x horixontal and y vertical)
+            # But our mazes and paths use array notation (row, col)
+            # So we swap the dimensions here
+            y: NDArray = p_transformed[:, 0]
+            x: NDArray = p_transformed[:, 1]
             plt.quiver(
                 x[:-1],
                 y[:-1],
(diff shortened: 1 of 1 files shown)


rusheb:
Amazing work!

afspies:
Looks good - for my own clarity, we are using (x<->columns, y<->rows) with (0,0) at the top?

Is this the same coordinate system as in the tokenization (i.e. (0,0)<->(0,1) => a connection from the top left to the cell underneath?

afspies (commented):
Looks good - for my own clarity, we are using (x<->columns, y<->rows) with (0,0) at the top left?

Is this the same coordinate system as in the tokenization (i.e. (0,0)<->(0,1) => a connection from the top left to the cell underneath?

mivanit (approved):
LGTM! this is also much cleaner than the cursed `rot90` haha

valedan:
> Looks good - for my own clarity, we are using (x<->columns, y<->rows) with (0,0) at the top left?

Yep!

> Is this the same coordinate system as in the tokenization (i.e. (0,0)<->(0,1) => a connection from the top left to the cell underneath?

In the tokenization we use (rows, cols), so (0,0)<->(0,1) is a connection from top left to the cell on the right (ie 1 col over).

But [plt.plot](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html) uses Cartesian (x, y) coordinates, so (0,0)<->(0,1) is a connection from top left to the cell underneath (moving down the y-axis). 

So when plotting mazes, we need to swap the coordinate dimensions when passing coordinate tokens into `plt.plot`.

rusheb:
> > Looks good - for my own clarity, we are using (x<->columns, y<->rows) with (0,0) at the top left?
> 
> Yep!

Oh! I'm confused by this. I think we want to use (x<->rows, y<-> columns), so the visualization maps to the connection list

e.g. 
```
[
  [[T, T],
   [F, F]],
  [[T, F],
   [F, F]]
]
```

will be plotted like 
```
N - N
|   | 
N   N
```


That's what we are doing right?

valedan:
@rusheb I just tested it out, and for the connection list you provided the plot is:

Screenshot 2023-03-14 at 9 01 19 AM

rusheb:
okay perfect! so x is rows, not columns. right? or am I still mixed up?

valedan:
Ah okay this is a little confusing! When we do `plt.imshow(img` we leave `img` in (row, col) format because `imshow` handles that fine. Everything I said above about swapping dimensions only applies to the `path` coordinates that we pass into `plt.plot`

afspies (approved):
LGTM!

PR #86, opened 2023-03-14
Fix plot_loss and add integration test

luciaquirke:


tests/integration/maze_transformer/generation/test_plot_loss.py
```diff
@@ -0,0 +1,6 @@
+from maze_transformer.evaluation.plot_loss import plot_loss
+
+
+def test_plot_loss():
+    path_from_root = "data/maze/g4-n100/g4-n100_tiny-v1_2023-03-13-11-55-48/log.jsonl"
+    plot_loss(f"../../../../{path_from_root}")
```
luciaquirke, line 6: lol I think I was looking for an absolute path here

maze_transformer/evaluation/plot_loss.py
```diff
@@ -52,6 +52,7 @@ def plot_loss(
     total_sequences, loss = zip(*data_raw)
 
     # compute a rolling average
+    # this messy part here is just for handling command line arguments
```
rusheb, line 55: Maybe worth extracting a method  called `parse_window_sizes` or something like that, then we wouldn't need this comment

tests/integration/maze_transformer/generation/test_plot_loss.py
```diff
@@ -0,0 +1,39 @@
...
+from maze_transformer.evaluation.plot_loss import plot_loss
+from scripts.create_dataset import create_dataset
+from scripts.train_model import train_model
+
+
+@pytest.fixture()
+def temp_dir() -> Path:
+    data_dir = tempfile.TemporaryDirectory()
+    yield Path(data_dir.name)
+    data_dir.cleanup()
```
rusheb, line 16: This fixture is used in a few different files now.

We could try moving this fixture to a new module `tests.integration.fixtures` and importing it.

rusheb (approved):
Thanks for this! Added some minor comments - both optional

tests/integration/maze_transformer/generation/test_plot_loss.py
rusheb: May require a bit of research to figure out the optimal way to do this. 

More info here: https://gist.github.com/peterhurford/09f7dcda0ab04b95c026c60fa49c2a68


We could move this into a separate ticket, but that seems to me like the kind of the ticket which would never get prioritised.

maze_transformer/evaluation/plot_loss.py
mivanit: good point!

tests/integration/maze_transformer/generation/test_plot_loss.py
luciaquirke: Great call, did the first step and added a comment for the second

PR #87, opened 2023-03-14
as_img() drawing thinner walls

canrager:
- added option to adjust wall thickness
- added option to show connections (is this relevant?)
- noted changes in demo notebook


rusheb:
Hi Can, thank you for the PR! Great work on the visualisation - it looks really good! Just pasting a screenshot below so others don't have to clone your branch to see it:
image

I have a few comments and questions about this below.

## Desired behaviour
I'm just thinking about the desired behaviour for this feature. I think we should make sure everyone is agreed on what we want before merging it. Here are a couple of questions that I think we should check with the team:
* Do we want to *replace* the old visualisation or add an option to view it both ways?
* Do we need `node_length` and `show_connections` to be configurable or can we just hardcode some sensible values? 

## Other notes
### Automated checks
* The automated checks have failed. Let me know if you need help addressing that

### Code organization
* My intuition is telling me that the functions in `plot_maze.py` should just accept a LatticeMaze and figure out how to plot it, including producing the array. In this case we would move the logic from `as_img` into `plot_maze.py` which would encapsulate all of the plotting logic in one place. Would be good to hear your/the team's thoughts on this suggestion.

### Docstring
* I'm not sure the example in the docstring is still necessary now that you have made the vis much more intuitive. Maybe we could delete that part of the docstring. Not sure. 

### Modifying node_length on the fly
* I'm having an issue in the notebook when I try to modify `node_length` on the fly. First I set `node_length` to 11:
image

then I set `node_length` to 6 and rerun the cell:

image

### Plot_path
* It doesn't seem to work with `plot_path` yet: 
image
* We should also check `plot_multi_paths`

valedan:
This looks great!

- I think it's fine to replace the old behaviour, the new version is much more intuitive IMO.
- I agree that the plotting logic should be put in `plot_maze.py`. The tick handling/imshow etc should be moved from the notebook into there. 
- Re: where should the logic live - I like the idea of keeping plotting logic in one place, so it makes sense to me to move all the `as_img` logic to `plot_maze`. BUT we currently have no way of testing `plot_maze` - I was thinking that we should extract most of the logic there into testable functions, and have the functions that actually call `imshow` or `plt.plot` be fairly small (since they're hard to test). 
- Re: snapshot testing - in my experience these are usually flaky and annoying. I also don't know if there is good tooling for these in the Python ecosystem. But I see why you're thinking of it - these plots may be hard to test the normal way. I'd suggest trying to write sensible tests without snapshots, and if it seems really hard then maybe explore snapshots.

valedan:
Style nitpick: I find the connections very distracting. Maybe they could be a bit lighter or thinner. Or is it just me?

rusheb:
> @rusheb @valedan is this wall thickness fine? Should connectins be shown at all (otherwise just coloring them white)? If yes, is the "lightness" of the gray fine?

It looks fine to me!

canrager:
As a next step, I will make coloring nodes with attention matrix available. Should this be done in a new issue? If not, leave this PR open.

rusheb:
> As a next step, I will make coloring nodes with attention matrix available. Should this be done in a new issue? If not, leave this PR open.

Yep, I think a new issue would be good. Best to try to keep PRs as small as possible.

rusheb:
@canrager it appears you have some merge conflicts. I think due to #81 being merged. 

Let me know if you want a hand with resolving them.

maze_transformer/evaluation/plot_maze.py
```diff
@@ -6,20 +6,113 @@
...
+
+     Axes definition:
+    (0,0)     y
+      -|----------->
+     x |
+       |
+       |
+       v
+
+     Returns a matrix of side length (ul) * n + 1 where n is the number of nodes.
```
rusheb, line 36: Nice description!

maze_transformer/evaluation/plot_maze.py
```diff
@@ -6,20 +6,113 @@
...
+    """
+
+    # Set color of connections (using plt.imshow(cmap='grey))
+    show_connections = True
+
+    white_color = 100
+    if show_connections:
+        connection_color = white_color * 0.93
+    else:
+        connection_color = white_color
```
rusheb, line 46: `show_connections` should either be a parameter or it should be removed. I say we remove it, then we can simplify this.

maze_transformer/evaluation/plot_maze.py
```diff
@@ -6,20 +6,113 @@
...
 
+CoordArray = NDArray["coords", np.int8]
+
+# Set size of unit.
+# Each "unit" consists of a node and the right and lower adjacent wall/connection. Its area is ul * ul.
+# Wall thickness is fixed to 1, so this parameter sets the ratio of node length to wall thickness
+ul = 14
+
+
+# Maze conversion functions (No plt involved)
```
rusheb, line 17: I guess you are doing this outside because as_img and points_transform_to_img both depend on it? 

In that case I think we should introduce a class.

We could have a nice [fluent interface](https://martinfowler.com/bliki/FluentInterface.html) like:
```
MazePlot(maze).add_path(path).add_path(path).show()
```
or
```
MazePlot(maze).add_multi_paths(path).show()
```


Let me know if you want to pair on this. Also let me know if i am over-engineering it! 😅

maze_transformer/evaluation/plot_maze.py
```diff
@@ -6,20 +6,113 @@
...
+
+    return img
+
+
+def points_transform_to_img(points: CoordArray) -> CoordArray:
+    """transform points to img coordinates"""
+
+    #####
+    # Swapping x, y axes here may be deprecated with resolution of issue #73
+    points = np.array([(x, y) for (y, x) in points])
```
rusheb, line 79: Hmm..yeah, we shouldn't need to do this any more. There's something fishy about this

valedan: You need to do this any time you're passing a list of `Coords` into `plt.plot` or similar functions, because the former use (row, col) array coordinates and the latter use (x, y) Cartesian coordinates, which are opposite.

In this case we definitely shouldn't be doing it - this function returns a `CoordArray` and those should always be (row, col)

afspies:
> As a next step, I will make coloring nodes with attention matrix available. Should this be done in a new issue? If not, leave this PR open.

Have created #108

maze_transformer/evaluation/plot_maze.py
canrager: thanks, added row col notation here

maze_transformer/evaluation/plot_maze.py
canrager: @afspies thats right, renamed to row / col as suggested

maze_transformer/evaluation/plot_maze.py
canrager: good suggestion, I implemented the fluent interface!

canrager:
All `matplotlib` code is called in `MazePlot.show()` `._plot_path()` and `._plot_maze()`. This makes testing other functions of the class easier. However, I lean towards not writing tests here. Researchers may customize visualizations and tests may be hard to maintain. Additionally, its most probable to notice a bug directly in the image, after some plotting function was broken by accident. A unit test would not add too much value. What do you think? (Discussing the reasoning cause I'd like to gain some intuition for testing.)

canrager:
Last test fails due to an error in `LatticeMaze.as_ascii()`: This function uses `.as_img()` which has been replaced by `evaluations.plot_maze.MazePlot._latticemaze_to_img()` in this PR. Do we need `as_ascii` or is it kinda duplicating the `plt` visualisation we already have? If we decide to keep `as_ascii`, I'd propose to move it to `evaluations.plot_maze.MazePlot`. What do you think?

afspies:
> 
I wanted as_ascii in case one is doing stuff through the terminal but wants some minimal debugging.

Happy to ditch it or move it into the new Class though.

canrager:
> I wanted as_ascii in case one is doing stuff through the terminal but wants some minimal debugging.

Added as ascii to the MergePlot class.

Screenshot 2023-03-24 at 18 39 51

mivanit:
This is not immediately urgent, but perhaps we could split `as_ascii()` into an `as_pixels()` function which returns a boolean array of shape `(3, n, n)`. 0th index would be wall/not wall, 1st and 2nd for start/target, respectively. I think this is worth it because we may in the future want to have the ability to feed in a rasterized maze, or see how our transformers compare to RCNNs trained on [similar tasks](https://arxiv.org/abs/2106.04537), which are quite easily interpretable. There might also be clever things to do if our residual stream is an embedding of an image.

the interface of `as_ascii()` would be unchanged, it would just be a relatively simple processing of the results of `as_pixels()`.

However, `as_pixels()` might not be the best name -- any suggestions?

maze_transformer/evaluation/plot_maze.py
```diff
@@ -31,44 +15,193 @@ class PathFormat:
...
+        """
+        Recieve predicted path and formatting preferences from input and save in predicted_path list.
+        Default formatting depends on nuber of paths already saved in predicted path list.
+        """
+        if label is None:
+            label = "predicted path {path_num}".format(
+                path_num=len(self.predicted_paths) + 1
+            )
+        if color is None:
+            color = self.DEFAULT_PREDICTED_PATH_COLORS[len(self.predicted_paths)]
```
rusheb, line 90: this will throw if len(predicted_paths) > len(DEFAULT_PREDICtED_PATH_COLOrS)

We should either:
* Fix with a modulo operation or `itertools.cycle`
* Raise a custom exception

maze_transformer/evaluation/plot_maze.py
```diff
@@ -31,44 +15,193 @@ class PathFormat:
...
+                if self.maze.connection_list[1, row, col]:
+                    img[
+                        row * unit_length + 1 : (row + 1) * unit_length,
+                        (col + 1) * unit_length,
+                    ] = connection_color
+
+        return img
+
+    def _rowcol_to_coord(self, points: CoordArray) -> NDArray:
+        """transform points to img coordinates"""
```
rusheb, line 163: nit: I think this could be more descriptive - e.g. "Transform points from MazeTransformer to matplotlib default coordinate system"

maze_transformer/evaluation/plot_maze.py
```diff
@@ -31,44 +15,193 @@ class PathFormat:
...
+            path=path,
+            label=label,
+            fmt=fmt,
+            color=color,
+            line_width=line_width,
+            quiver_kwargs=quiver_kwargs,
+        )
+        return self
+
+    def add_predicted_path(
```
rusheb, line 72: Do we want an additional function that takes a collection of paths and adds them all? 

Feel free to reject this or raise a separate issue if it's out of scope

maze_transformer/evaluation/plot_maze.py
```diff
@@ -31,44 +15,193 @@ class PathFormat:
...
+                    maze_str += wall_char
+            maze_str += "\n"  # Start a new line after each row
+        return maze_str
+
+    # Image rendering functions (plt involved)
+    def _plot_maze(self) -> None:
+        img = self._latticemaze_to_img()
+        self.ax.imshow(img, cmap="gray", vmin=0, vmax=1)
+
+    def _plot_path(self, pf: PathFormat) -> None:
```
rusheb, line 199: nit: rename pf to path_format

rusheb (changes requested):
Almost there! The most important comment is about the error behaviour once we exhaust the DEFAULT_PREDICTED_PATH_COLORS.

Also, please double check the notebooks as I can see that `eval_model.ipynb` and `plot_attention.ipynb` are still importing/using the old plotting functions

Rest of the comments are minor/optional

maze_transformer/evaluation/plot_maze.py
canrager: added modulo as it requires minimal changes

maze_transformer/evaluation/plot_maze.py
canrager: true, now: Transform Points from MazeTransformer (row, column) notation to matplotlib default (x, y) notation where x is the horizontal axis.

rusheb (approved):
great work 🚀

PR #94, opened 2023-03-15
#88 - random paths

### LINKED ISSUE #88: Each maze should use a random path, not the path from top-left to bottom-right
Acceptance criteria:
??

valedan:
Resolves #88 
 
Changes:
- `create_dataset` and `test_generation` scripts will now generate maze tokenizations with random paths 
- Tiny update to notebook setup instructions

To test:
- Create a new dataset
- Run `eval_model` notebook with the new dataset. You should see that the paths are now random in the plotted mazes

Screenshot 2023-03-15 at 11 47 26 AM


maze_transformer/generation/latticemaze.py
```diff
@@ -290,6 +291,24 @@ def find_shortest_path(
...
+            (row, col)
+            for row in range(self.grid_shape[0])
+            for col in range(self.grid_shape[1])
+        ]
+
+    def generate_random_path(self) -> list[Coord]:
+        """ "return a path between randomly chosen start and end nodes"""
+
+        # we can't create a "path" in a single-node maze
+        assert self.grid_shape[0] > 1 and self.grid_shape[1] > 1
```
rusheb, line 307: nit: should this be an `or` condition rather than `and`?

maze_transformer/generation/latticemaze.py
```diff
@@ -290,6 +291,24 @@ def find_shortest_path(
...
                 f_score[neighbor] = g_score[neighbor] + self.heuristic(neighbor, c_end)
 
+    def get_nodes(self) -> list[Coord]:
+        """return a list of all nodes in the maze"""
+
+        return [
+            (row, col)
+            for row in range(self.grid_shape[0])
+            for col in range(self.grid_shape[1])
+        ]
```
rusheb, line 301: nit: does order matter? If not, should this be a set?

rusheb (approved):
Great stuff! I ran through the manual testing steps you suggested and the paths look sensible to me

maze_transformer/generation/latticemaze.py
valedan: I think you're right, order does not matter here. I think if we ever care about order in future, we would probably want a seperate method that returns a 2D array of nodes.

valedan: Okay I tried using a set and discovered that sampling from a set is deprecated in python 3.9 and removed as of 3.11. So this actually needs to be a list.

scripts/create_dataset.py
```diff
@@ -15,8 +15,6 @@
 def generate_MazeTokenizer(
     junk,
```
afspies, line 16: What is going on with this junk parameter?

afspies (approved):
LGTM

scripts/create_dataset.py
valedan: Either `pool.imap` or `tdqm` adds a first parameter to the wrapped function that tracks something like iteration or thread number or something like that. We're not using it.

PR #100, opened 2023-03-18
Fix array out of bounds error in IndexedArray and add tests

luciaquirke:
Class contains an array made by concatenating several arrays, and another array containing the starting index of each original array in the concatenated array. Later uses the starting point of the next array to find the length of each array, which throws an out of bounds exception for the final one.

Another solution is to remove the [:-1] from `torch.cumsum(torch.tensor([0, *map(len, data)]), dim=0)[:-1]`, so `i + 1` always exists for valid `i`s, but it makes the data structure contain a weird edge case value that should never be directly accessed. In the long term I think the best solution might be to see if there is a more natural data structure we could use that still plays nicely with `__getitem__` with datasets.

### DIFF
--- a/maze_transformer/training/training.py
+++ b/maze_transformer/training/training.py
@@ -143,7 +143,6 @@ def train(
     n_batches: int = len(dataloader)
     logger.log({"n_batches": n_batches}, 10)
 
-    n_sequences: int
     print_loss_interval_iters: int = int(
         cfg.train_cfg.print_loss_interval // cfg.train_cfg.batch_size
     )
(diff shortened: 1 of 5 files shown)


maze_transformer/training/training.py
```diff
@@ -152,7 +151,7 @@ def train(
     )
     for iteration, batch in enumerate(dataloader):
         # compute loss
-        with TimerContext() as timer_loss:
+        with TimerContext() as loss_time:
```
luciaquirke, line 154: switched this around bc it confused me the first time I read it

valedan (approved):
Nice fix! 👍

Was this the cause of the sequence truncation during training or is this a separate bug you discovered?

PR #101, opened 2023-03-18
Rename idxs/idx to index/i/indices

luciaquirke:
- I know this is a bit psycho
- index/indices is slightly more conventional and I read it faster


valedan (approved):
I don't have a strong preference either way here, but seems reasonable! 😄

tests/unit/maze_transformer/training/test_dataset.py
```diff
@@ -0,0 +1,35 @@
+import torch
```
valedan, line 1: Did you mean to include this file or was it part of your other PR?

luciaquirke: Nope, cheers!

rusheb:
> I know this is a bit psycho

Small readability PRs like this are awesome imo - keep them coming! 🦀

PR #102, opened 2023-03-19
Break SolvedMaze out of MazeTokenizer

valedan:
Some relatively minor refactoring.

I noticed that `MazeTokenizer` had 2 distinct roles:

1. Store a maze and a solution together
2. Tokenize a maze

I think this caused some confusion in the code in a few places.

I've broken the first role out into a separate `SolvedMaze` class, which is just a NamedTuple that stores a maze and a solution. This leaves just the tokenization function behind in `tokenizer.py`, which should pave the way to move it into @afspies new Tokenizer when that gets merged. Doing this let me simplify things in a few other places, especially the create_dataset script.

I decided to go with a new class rather than putting a `solution` attr directly on `LatticeMaze` for a few reasons:
- Conceptually mazes and solutions are distinct
- I'm not confident that it will always make sense for mazes to only store one solution, so I don't want to bake that into the main class.
- If it was on the class, then when you're working with a `LatticeMaze` object in a notebook, you would never be quite sure if it contained a solution or not. With a `SolvedMaze` you always know it does.

### DIFF
--- a/maze_transformer/generation/generators.py
+++ b/maze_transformer/generation/generators.py
@@ -3,7 +3,12 @@ from typing import Any, Callable
 
 import numpy as np
 
-from maze_transformer.generation.latticemaze import NEIGHBORS_MASK, Coord, LatticeMaze
+from maze_transformer.generation.latticemaze import (
+    NEIGHBORS_MASK,
+    Coord,
+    LatticeMaze,
+    SolvedMaze,
+)
 
 
 class LatticeMazeGenerators:
@@ -14,7 +19,7 @@ class LatticeMazeGenerators:
         grid_shape: Coord,
         start_coord: Coord | None = None,
         lattice_dim: int = 2,
-    ) -> "LatticeMaze":
+    ) -> LatticeMaze:
         """generate a lattice maze using depth first search, iterative
 
         algorithm:
@@ -93,6 +98,13 @@ class LatticeMazeGenerators:
                 start_coord=start_coord,
             ),
         )
+
+    @classmethod
+    def gen_dfs_with_solution(cls, grid_shape: Coord):
+        maze = cls.gen_dfs(grid_shape)
+        solution = np.array(maze.generate_random_path())
+
+        return SolvedMaze(maze, solution)
 
 
 GENERATORS_MAP: dict[str, Callable[[Coord, Any], "LatticeMaze"]] = {
--- a/maze_transformer/generation/latticemaze.py
+++ b/maze_transformer/generation/latticemaze.py
@@ -1,5 +1,6 @@
 import random
 from dataclasses import dataclass
+from typing import NamedTuple
 
 import numpy as np
 from muutils.misc import list_split
@@ -382,3 +383,10 @@ class LatticeMaze:
             adjlist[i, 1] = np.array(coord_str_to_tuple(c_end))
 
         return cls.from_adjlist(adjlist)
+
+
+class SolvedMaze(NamedTuple):
+    """Stores a maze and a solution"""
+
+    maze: LatticeMaze
+    solution: list[tuple[int, int]]
--- a/maze_transformer/training/tokenizer.py
+++ b/maze_transformer/training/tokenizer.py
@@ -1,9 +1,8 @@
-from dataclasses import dataclass, field
 from itertools import chain
 
 from maze_transformer.generation.latticemaze import (
     SPECIAL_TOKENS,
-    CoordArray,
+    Coord,
     CoordTup,
     LatticeMaze,
 )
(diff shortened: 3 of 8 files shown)


maze_transformer/generation/latticemaze.py
```diff
@@ -382,3 +383,10 @@ def from_tokens(cls, maze_tokens: list[str]) -> "LatticeMaze":
             adjlist[i, 1] = np.array(coord_str_to_tuple(c_end))
 
         return cls.from_adjlist(adjlist)
+
+
+class SolvedMaze(NamedTuple):
+    """Stores a maze and a solution"""
```
luciaquirke, line 389: 🔥 🔥 🔥

maze_transformer/training/tokenizer.py
```diff
@@ -1,66 +1,47 @@
...
+        SPECIAL_TOKENS["origin_start"],
+        node_token_map[tuple(solution[0])],
+        SPECIAL_TOKENS["origin_end"],
+        # give target
+        SPECIAL_TOKENS["target_start"],
+        node_token_map[tuple(solution[-1])],
+        SPECIAL_TOKENS["target_end"],
+        SPECIAL_TOKENS["path_start"],
+        *[node_token_map[tuple(c.tolist())] for c in solution],
+        SPECIAL_TOKENS["path_end"],
```
rusheb, line 44: ⚠️ Will this modify the tokenization scheme such that "path_start" and "path_end" tokens are always included, regardless of whether the solution is empty? 

This might be a change for the better, but I don't think it belongs in a refactor. At the very least we should check with the team that this behaviour is desired.

rusheb (changes requested):
Definitely a move in the right direction! 

One potential change in behaviour that I'm concerned about. I think if we had more tests in place we could be more confident about not changing behaviour when refactoring -- however I think this is a really helpful change, and I don't want to increase the barrier to doing stuff like this. So I'm not sure how to strike a balance between adding tests and doing helpful refactorings.

notebooks/plot_attention.ipynb
```diff
@@ -1,1115 +1,1110 @@
...
+                "path_true = np.array(maze.find_shortest_path(\n",
+                "\tc_start = c_start,\n",
+                "\tc_end = c_end,\n",
+                "))\n",
+                "\n",
+                "\n",
+                "# tokenize the maze\n",
+                "tokens = maze_to_tokens(maze, path_true, cfg.dataset_cfg.node_token_map)\n",
+                "path_start_index = tokens.index(SPECIAL_TOKENS[\"path_start\"])\n",
+                "maze_only_tokens = tokens[:path_start_index + 1]\n",
```
rusheb, line 92: It's interesting that we don't bother using the SolvedMaze here. It seems like at any point we have a maze and solution we could call `maze_to_tokens` or `solved_maze_to_tokens`. I think this is confusing.

Which made me think, maybe we should introduce a `MazeTokenizer` class which is responsible for tokenizing a maze with or without a soluition.....uh oh.

Maybe it would be sensible to create overloads of `maze_to_tokens` which take, respectively, [maze], [maze, solution], and [SolvedMaze]. But I don't think that is easy to do in python. So maybe we could use flags? [multipledispatch](https://stackoverflow.com/a/29091980/6631502) may help us here, but haven't looked in detail.

(Maybe this is more of a minor issue than I am making out)

tests/unit/maze_transformer/training/test_maze_dataset.py
```diff
@@ -3,36 +3,36 @@
...
-from maze_transformer.training.tokenizer import MazeTokenizer
+from maze_transformer.generation.latticemaze import LatticeMaze, SolvedMaze
+from maze_transformer.training.mazedataset import (
+    MazeDatasetConfig,
+    solved_maze_to_tokens,
+)
 
 
-def test_maze_to_tokens():
+def test_solved_maze_to_tokens():
```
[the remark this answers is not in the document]
valedan: No, `solution` is a required parameter.

maze_transformer/training/tokenizer.py
valedan: Yes, with this change solution is a required parameter and will always be included. I wouldn't call this a change to the tokenization scheme though, everything gets tokenized the same way.

This change actually has no functional effect, it just simplifies things a lot. In the [previous version](https://github.com/AISC-understanding-search/maze-transformer/blob/main/maze_transformer/training/tokenizer.py), we needed to construct a `MazeTokenizer` and then call `as_tokens`. `solution` was a required property when constructing the tokenizer, and then when calling `as_tokens` you could specify whether to include it in the result or not. But when actually using the function, we always included the solution, except in `plot_attention.ipynb`, which I've adjusted to maintain the old behaviour.

I think it's overall simpler for the tokenizer to not have too much logic and just return the full tokenization. In my evals branch I've added a bunch of [utility functions](https://github.com/AISC-understanding-search/maze-transformer/blob/99-wip-2/maze_transformer/utils/token_utils.py) to easily get different parts of the tokenization, including one that gets everything up to but not including the solution, which is equivalent to the old behaviour here when the solution is excluded.

maze_transformer/generation/generators.py
```diff
@@ -94,6 +99,13 @@ def gen_dfs(
             ),
         )
 
+    @classmethod
+    def gen_dfs_with_solution(cls, grid_shape: Coord):
```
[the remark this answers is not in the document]
valedan: Yep good call!

maze_transformer/generation/generators.py
```diff
@@ -94,6 +99,13 @@ def gen_dfs(
...
+        maze = cls.gen_dfs(grid_shape)
+        solution = np.array(maze.generate_random_path())
```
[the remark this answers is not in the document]
valedan: Yep in DFS these are the same. But good point that this will become confusing when we add new maze gen algorithms. I'll add a comment about this to the relevant issue.

notebooks/plot_attention.ipynb
valedan: I think the problem here is that `solved_maze_to_tokens` shouldn't exist. It's really more of an internal method to `MazeDataset` to allow easier use of partials, but it's not really needed and I agree it causes confusion. I'll see if it's straightforward to remove as part of this PR.

I agree that a future enhancement could be to allow tokenizing mazes that don't have a solution, but I'm not sure about the best approach there and I think some research/thinking will be needed. It's not a very urgent use case (this is the only part of the codebase that we would use it currently), so I'd prefer to make a ticket for it and do it later.

valedan: Okay, removed `solved_maze_to_tokens`.

maze_transformer/training/tokenizer.py
afspies: Big fan of this change as the behaviour w.r.t `solution` caused me some annoyance in trying to include `LatticeMazes` In the `WrappedTokenizer` encoding method

rusheb (approved):

maze_transformer/generation/generators.py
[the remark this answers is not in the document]
luciaquirke: `generate_random_path` uses the [A* algorithm](https://en.wikipedia.org/wiki/A*_search_algorithm), so if it's implemented correctly it will always find the shortest path from the specified source to the specified goal. The docstring for `generate_random_path` could be updated to clarify this.

Issue #35, opened 2023-02-20
Fix warning when running integration test


rusheb:
The integration test produces the following warning:

> tests/integration/test_train_model.py::test_train
>   /Users/rusheb/code/maze-transformer/venv/lib/python3.10/site-packages/torch/utils/data/dataloader.py:554: UserWarning: This DataLoader will create 16 worker processes in total. Our suggested max number of worker in current system is 8 (`cpuset` is not taken into account), which is smaller than what this DataLoader is going to create. Please be aware that excessive worker creation might get DataLoader running slow or even freeze, lower the worker number to avoid potential slowness/freeze if necessary.
>     warnings.warn(_create_warning_msg(
> 
> Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
> 

We should fix it by reducing num_workers in the training config

cmathw:
Addressed in this PR: change num_workers #103

rusheb:
Be my guest! Thanks for the fix!

PR #103, opened 2023-03-19
change num_workers

cmathw:
Addresses minor issue in #35 

Minor issue with num_workers used in training/config.py (changed tiny-v1 num_workers from 16 to 8, this config is used by integration/test_train_model.py).


luciaquirke (approved):
LGTM! 🚀

PR #105, opened 2023-03-22
Fix layernorm folding issue when loading from a trained model which d…

afspies:
Fixes issues around layernorm hooks being broken with our current model loading process. 

This is a temporary fix until https://github.com/neelnanda-io/TransformerLens/issues/219 is resolved


rusheb:
Changes look good! 

Just to clarify - will this break once https://github.com/neelnanda-io/TransformerLens/issues/219 is resolved? 

Either way, I think it would be good to add a code comment referencing the issue.

rusheb (approved):
🚀

cmathw:
@afspies @rusheb I think we are still having this issue? At least when I retrain a model with this change, run_with_cache and check the cache_dict no layer norms related activations are present. It still seems to be having like a model with normalization_type = None (as opposed to "LNPre").

cmathw:
I have been pretty confused about this because our model seems to be using normalization_type = LNPre and fold_ln = True

I have tried to make sense of this here (to little avail): https://github.com/AISC-understanding-search/experiments/blob/main/chris/layer_norm_confusion.ipynb

PR #110, opened 2023-03-23
Weights & Biases integration

rusheb:
Not super happy with this, but I've been staring at it for too long!

I've tried to do a minimal implementation which we can extend according to priorities. 

It would be good if reviewers could run the `train_model` script manually and check that there's no important information that I have deleted, and also maybe check if there's any unimportant information that I should delete.

Example run: https://wandb.ai/aisc-search/rusheb-testing/runs/gpyqu6zn

# Changes
* Wrap w&b API with new `WandbLogger`class.
* Modify existing logging to use w&b
  * loss is uploaded using the `log`method because it is a metric that changes throughout the run
  * most other items are logged using `summary` because they are constant in the context of a run
  * progress update logs (e.g. "Initializing optimizer") are just logged to the console, because the whole of stdout is captured and uploaded to w&b
  * final model and checkpoints are uploaded as artifacts
  * timers and memory usage are removed because w&b handles a lot of this for us
  * DELETED `plot_loss`-- because w&b plots the loss for every run and we are no longer saving the loss locally
* Integration tests will also log data to w&b, to a separate project space named "integration-tests" TODO Link
  * Annoyingly we can't get a service account on the academic tier, so i gave github actions my personal API key. I might create a dummy account to create some separation.
* `train_model.py` requires an extra argument: "wandb-project", which defines which project to upload the data to?

# Questions
* Not sure what the interface to `train_model.py` should be. 
  * Does the new required argument "wandb-project"  seem correct? 
  * Should it have a default? What should be the default behaviour?
    * Don't log to w&b
    * Log to a default project (which one?)
* I'm not really sure what meaningful tests I could add for the new WandbLogger class

# Future work
* Control how often the loss is logged to w&b in accordance to their [metric frequency guidelines](https://docs.wandb.ai/guides/track/limits#metric-frequency)
* Log the loss to a file locally
* There is an object "dataset_seq_len_stats" which is being uploaded to the summary. This should be converted to a graph to show on the dashboard.
* Support resuming runs from a checkpoint
* Add handlers to `WandbLogger` to help separate responsibilities of logging to w&b and to console. This would also allow us to add file handlers etc 
* Use an instance of `Logger` rather than the root logger to log to console from the `WandbLogger`
* Remove clutter from the summary to make it more useful
* Make "wandb-project" argument optional? add default?


luciaquirke:
I've played around with the commands in the README and have gotten two errors:
The mysterious
image
And the more mundane
image
Not exactly sure what the problem is

rusheb:
@lucia thank you for trying this out!

I think the second error might be due to an old dataset. When did you generate the dataset? If you generate a new one and run the train_model code do you still get the error? 

I'm not sure about the first error. I can't seem to reproduce it.

luciaquirke:
That was it, thank you! That directory contained several datasets with different timestamps, I guess the cfg.json file isn't updated when a new dataset is added 🤔

maze_transformer/training/training.py
```diff
@@ -35,175 +33,85 @@ class TRAIN_SAVE_FILES:
...
-    dataset: MazeDataset, cfg: ConfigHolder, logger: Logger
+    dataset: MazeDataset, cfg: ConfigHolder, logger: WandbLogger
 ) -> DataLoader:
-    logger.log_elapsed_last()
-    logger.mem_usage()
-
     length_stats: StatCounter = StatCounter(dataset.get_all_lengths())
-    logger.log({"dataset_seq_len_stats": length_stats.summary()})
-    logger.log({"dataset_seq_len_stats": length_stats.serialize()}, lvl=50)
+    logger.summary({"dataset_seq_len_stats_summary": length_stats.summary()})
```
luciaquirke, line 40: image
Looks like the serialized version below already includes the summary?

luciaquirke (approved):
Huge fan of the amount of code removed here. I've approved this because I can't see any major issues and people keep talking about their exotic loss curves and I want access to them asap.

luciaquirke:
I think not logging to W&B should be the default behaviour, for integration tests and users without W&B set up.

Q: do we need plot_loss and other hand rolled plotting methods for writing blog posts etc.? How do people usually handle that?

maze_transformer/training/training.py
rusheb: Yes it does. I think for now I might just log the summary to the console so it doesn't clutter up the wandb summary

PR #120, opened 2023-03-24
[WIP] Suppress warning from setuptools

luciaquirke:
Suppress DeprecationWarning coming from setuptools

From https://github.com/spdx/tools-python/issues/507 it looks like it may be coming from the `__import__("pkg_resources").declare_namespace(__name__)` line in isort but I don't know how to diagnose these properly 🤔


rusheb:
Thanks! Just wondering, is this issue definitely not something we can fix on our end? The issue https://github.com/pypa/setuptools/issues/2466 that you link to in the comment is closed. What exactly is causing this warning?

PR #131, opened 2023-03-26
Remove extra log in line with rusheb discussion

luciaquirke:
Smol fixup

### DIFF
--- a/maze_transformer/training/training.py
+++ b/maze_transformer/training/training.py
@@ -37,7 +37,6 @@ def get_dataloader(
     dataset: MazeDataset, cfg: ConfigHolder, logger: WandbLogger
 ) -> DataLoader:
     length_stats: StatCounter = StatCounter(dataset.get_all_lengths())
-    logger.summary({"dataset_seq_len_stats_summary": length_stats.summary()})
     logger.summary(
         {"dataset_seq_len_stats": length_stats.serialize(typecast=lambda x: str(x))}
     )


maze_transformer/training/training.py
```diff
@@ -37,7 +37,6 @@ def get_dataloader(
     dataset: MazeDataset, cfg: ConfigHolder, logger: WandbLogger
 ) -> DataLoader:
     length_stats: StatCounter = StatCounter(dataset.get_all_lengths())
-    logger.summary({"dataset_seq_len_stats_summary": length_stats.summary()})
     logger.summary(
```
rusheb, line 40:
```suggestion
    logger.progress(
```


I reckon we should move it from the summary to a console log, because it clutters up the summary 

Side-note, I'm kind of thinking the WandbLogger interface is a bit obscure. Maybe the `progress` function should be called something different?

luciaquirke: Ah yeah I didn't catch the meaning of `progress`. `console`?

luciaquirke: I think we should update to a general W&B class that you pass a logger into, rather than a logger class that does other W&B operations like uploading artifacts. I can put up a draft if that sounds good to you

luciaquirke: Like 

```
W&BClass:
 def init(logger: CustomLogger):
   # etc
 def upload_dataset:
   # artifact = upload_dataset operation
   logger.wandb.log_artifact(artifact)
    
   

``

Comment (luciaquirke) [reply maze_transformer/training/training.py:41]: Hmm I need to think more about this

Comment (luciaquirke) [reply maze_transformer/training/training.py:41]: Does the 

```
        logging.basicConfig(
            stream=sys.stdout,
            level=logging.INFO,
            format="%(asctime)s %(levelname)s %(message)s",
            datefmt="%Y-%m-%d %H:%M:%S",
        )
```

code affect the W&B logging in any way? Maybe we could extract out the regular logger to another file and then call `logger` or `wandb` with its own logging encapsulated as necessary.

rusheb: That makes sense, happy to do that. Thanks for looking into it!

luciaquirke: I've moved it to the utils module for now instead of its own setup class to avoid over-engineering

luciaquirke:
I assume this is failing due to muutils error, still keen for review

rusheb:
@luciaquirke would you still like to try and merge this or should we close it?

PR #132, opened 2023-03-26
Use underscores to separate words in file and variable names

luciaquirke:
Small code health update created by case sensitive global find and replace for adjlist and ADJLIST, and manual scans through our file names

- updates tokenizer so old mazes may not work as well on new datasets and vice versa

### DIFF
--- a/maze_transformer/generation/lattice_maze.py
+++ b/maze_transformer/generation/lattice_maze.py
@@ -15,7 +15,7 @@ from maze_transformer.generation.constants import (
 )
 from maze_transformer.utils.token_utils import (
     decode_maze_tokens_to_coords,
-    get_adjlist_tokens,
+    get_adj_list_tokens,
     get_path_tokens,
 )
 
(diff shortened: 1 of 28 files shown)


maze_transformer/generation/lattice_maze.py
```diff
@@ -11,8 +11,8 @@
 # 	n_nodes: int
 
 SPECIAL_TOKENS: dict[str, str] = dict(
-    adjlist_start="<ADJLIST_START>",
```
valedan, line 14: I'm wary of changing the tokenization scheme, it seems like it will cause a headache for minimal benefit. Everything else in this PR seems fine, but could we leave ADJLIST_START and ADJLIST_END alone?

luciaquirke: Good call, done

mivanit:
Not an official reviewer, but this looks good to me provided merge conflicts are handled.

valedan (approved):
Ship it 🚀🚀🚀

PR #141, opened 2023-03-28
MazePlot custom node coloring

canrager:


maze_transformer/evaluation/plot_maze.py
```diff
@@ -116,9 +122,24 @@ def add_multiple_paths(self, path_list: list):
                 self.add_predicted_path(path=path)
         return self
 
-    def show(self) -> None:
+    def add_node_values(self, node_values: NDArray) -> MazePlot:
```
mivanit, line 125: this is very minor, but you shouldn't worry about adding `NDArray` or `ATensor` annotations -- use jaxtyping instead. Fox example, this could be
`jaxtyping.Float[np.ndarray, "grid_n grid_n"`

maze_transformer/evaluation/plot_maze.py
```diff
@@ -140,41 +161,72 @@ def show(self) -> None:
...
-
-    def _latticemaze_to_img(
-        self, unit_length: int = UNIT_LENGTH
-    ) -> NDArray["row col", bool]:
+        if self.custom_node_value_flag is False:
+            self.ax.imshow(img, cmap="gray", vmin=-1, vmax=1)
+
+        else:  # if custom node_values have been passed
+            n_blues = int(256 * self.max_node_value)  # for scaling colorbar up to value
+            blues = mpl.colormaps["Blues"].resampled(n_blues)  # load blue spectrum
```
mivanit, line 180: Is there a reason for this to always be blue?

canrager: Do we need an option for the user to plug other color spectra here? I assumed this is a choice we do once and it'll be a good fit for all experiments. `Blues` is good as we need a color spectrum from dark to light, which does go to black (to have the walls always visible).

maze_transformer/evaluation/plot_maze.py
canrager: as discussed, leaving it here as is, implementing jaxtyping from now on.

maze_transformer/evaluation/plot_maze.py
canrager: resolved
Screenshot 2023-03-29 at 17 08 54

mivanit: my reasoning from slack:

I imagine that we will want to display attention over the maze, and usually you use different colors for different heads. Being able to make the colors match would be useful for if we have a set of mazes with attention heatmaps for each and a set of actual attention matrices for each head.

Thanks for adding this!

maze_transformer/evaluation/plot_maze.py
```diff
@@ -116,9 +123,27 @@ def add_multiple_paths(self, path_list: list):
                 self.add_predicted_path(path=path)
         return self
 
-    def show(self) -> None:
+    def add_node_values(
+        self, node_values: NDArray, color_map: str = "Blues"
```
mivanit, line 127: it feels like this should be either a parameter to the `__init__` or to `add_node_values`, but not both. Did I misread this when I made my original comment?

maze_transformer/evaluation/plot_maze.py
```diff
@@ -24,28 +27,32 @@ class PathFormat:
...
+        "tab:purple",
+        "slategrey",
     ]
 
     def __init__(self, maze: LatticeMaze) -> None:
+        """
+        UNIT_LENGTH: Set ratio between node size and wall thickness in image.
+        Wall thickness is fixed to 1px
+        A "unit" consists of a single node and the right and lower connection/wall.
+        Example: ul = 14 yields 13:1 ratio between node size and wall thickness
```
mivanit, line 46:
```suggestion
        Example: ul = 14 yields 13:1 ratio between node size and wall thickness
        
        You can find valid colormaps here: https://matplotlib.org/stable/tutorials/colors/colormaps.html
```

canrager: thanks

maze_transformer/evaluation/plot_maze.py
canrager: My logic for the `add_node_values` function is it takes all relevant arguments for node highlighting, including the colormap. However other plotting functions within `MazePlot` class need to access the colormap parameter, thats why I initialize the class variable `self.node_color_map` in the `__init__` function. Should I initialize the `self.node_color_map` variable in another place than inside `__init__` function?

mivanit (approved):
Great work on this! @canrager let me know if the new `PlotFormat` interface makes sense, and then let's merge

maze_transformer/evaluation/plot_maze.py
mivanit: this makes sense now! I think the current interface is good.

Issue #122, opened 2023-03-24
Add Wilson's maze generation algorithm


luciaquirke:
Our mazes are generated using randomised depth first search. This generates mazes with a
[low branching factor and many long corridors](https://en.wikipedia.org/wiki/Maze_generation_algorithm#:~:text=Mazes%20generated%20with%20a%20depth,along%20each%20branch%20before%20backtracking.), because the algorithm explores as far as possible along each branch 
before backtracking.

Wilson's algorithm is an alternative which generates an unbiased sample of mazes from the uniform distribution over all
mazes, using loop-erased random walks. Shifting to mazes generated with this algorithm seems like it would help the 
model learn more generalisable search algorithms because the model would not learn to expect long corridors or other
maze-specific features.

valedan:
FYI the `generate_random_path` function might become slightly confusing once we have new generation algorithms, as described [here](https://github.com/AISC-understanding-search/maze-transformer/pull/102#discussion_r1142995253). We'll need to give some thought to this when implementing.

traeuker:
[Seems done](https://github.com/AISC-understanding-search/maze-transformer/pull/152)

PR #143, opened 2023-03-29
Create baseline model

valedan:
Add a baseline "model". The model has a `generate` function which has a compatible interface with `HookedTransformer.generate`, meaning it can be passed into `evaluate_model` just fine. 

The model uses a very simple algorithm to generate predictions: at each step it randomly selects an unvisited neighbouring node until it finds the target or reaches a dead end.

Not yet implemented - accepting a parameter to control the randomness - ie a custom weight for the correct path.

Demo notebook [here](https://github.com/AISC-understanding-search/experiments/blob/main/dan/00_baseline_evals.ipynb).

Example paths:
Screenshot 2023-03-29 at 11 39 33 AM

## Evaluations
Running `evaluate_model` on a `RandomBaseline` using a 10k maze dataset gives the following results:

----node_overlap----
Mean: 0.5930349117549116
Median: 0.5555555555555556

----num_connections_adjacent_lattice----
Mean: 4.2891
Median: 4.0

----fraction_connections_adjacent_lattice----
Mean: 0.7553681349206349
Median: 0.8

----num_connections_adjacent----
Mean: 4.2891
Median: 4.0

----fraction_connections_adjacent----
Mean: 0.7553681349206349
Median: 0.8


valedan:
Leaving as draft for now because I still need to tidy up the code a bit

mivanit:
re: accepting a parameter to control the randomness

Worth remembering here that both 2-forks and 3-forks should in theory be possible, and its not clear to me how this parameter should work in those contexts. setting a "probability of correct path" directly would bias it to be more correct at 3-forks, while multiplying `p` by the probability of randomly choosing the correct fork at that point also feels wrong since `p=2` would be fine for 2-forks and nonsensical for 3-forks. Maybe hit it with a sigmoid or something?

This feels like a pretty minor detail which could add a lot of complexity, not just here but in our analyses in general. Maybe we should add a maze generation algorithm (related to #122 ) which intentionally makes each node have at most 3 neighbors?

(as a side note, it might be interesting to see how a transformer trained on mazes with only 2-forks behaves when it encounters a 3-fork)

valedan:
Good points @mivanit, I'll leave that alone for now and see if we encounter a real need for a bias towards correct paths. 

As we experiment with trained models we may form hypotheses about what heuristics the model is using - it would be interesting to create additional baselines using those heuristics for comparison when that happens.

mivanit:
> Good points @mivanit, I'll leave that alone for now and see if we encounter a real need for a bias towards correct paths.
> 
> As we experiment with trained models we may form hypotheses about what heuristics the model is using - it would be interesting to create additional baselines using those heuristics for comparison when that happens.

from discussion today:

I think the best solution is to pass a parameter `p`, and then have the probability of taking an incorrect path be `base_prob * (1-p)`, where `base_prob` is `(n-1)/n` at a fork with `n` options. Then, at `p=1` we always take the correct path and at `p=0` we are randomly selecting.

valedan:
Closing in favour of #156 - that has the bias param, and a much cleaned up git history.

PR #147, opened 2023-03-30
Revert "Zanj integration for configs"

valedan:
Reverts AISC-understanding-search/maze-transformer#139

#139 broke loading of models created before the merge


mivanit:
fixed in #148 , closing

PR #149, opened 2023-03-31
Fixup indices

luciaquirke:
I can't see how this happened because there is no git blame in Jupyter notebooks 🤔 🤔 

I assume someone there was some race condition that could have been resolved by a rebase before merging

Need to check if we're running tests on main merges to catch this

Going to yolo merge because this concept has already been reviewed


mivanit (approved):
LGTM

PR #151, opened 2023-04-01
Add local maze to repo to ensure we always have one to run the notebooks with

luciaquirke:
Will later replace with one of the good mazes once w&b logs config files somewhere easily accessible


mivanit (approved):
LGTM! Will probably change this to a ZANJ after that is merged.

Also, imo it makes sense to have a minimal model in the repo, even after we have a "real" one in the wandb database

PR #152, opened 2023-04-02
Add Wilson's maze generation algorithm for unbiased maze datasets

luciaquirke:
- Will add maze generation with solution method in a follow-up PR - we can use the same A* implementation that the corresponding DFS method uses.
- [x] Need to add meaningful tests or visualise and verify manually, open to ideas on how to test well.
- [x] Need to document more properties of the generated mazes, specifically that they're acyclic


maze_transformer/generation/latticemaze.py
```diff
@@ -107,13 +93,16 @@ class LatticeMaze:
...
     """
+
     connection_list: NDArray["lattice_dim x y", bool]
     generation_meta: dict | None = None
 
     grid_shape = property(lambda self: self.connection_list.shape[1:])
 
     n_connections = property(lambda self: self.connection_list.sum())
 
+    lattice_dim: int = 2
```
luciaquirke, line 104: Old location for lattice_dim had an error because it has a default value so it needs to be after all the parameters without default values

scripts/test_generation.py
```diff
@@ -29,7 +29,7 @@ def generate_solve_plot(
 
     print(f"solving time: {time.time() - solution_start}")
 
-    plot_path(maze, path, show=True)
+    MazePlot._plot_path(maze, path, show=True)
```
luciaquirke, line 32: I'm not sure why this method is private now but it was throwing an error so I fixed

maze_transformer/generation/latticemaze.py
mivanit: `kw_only` is set to true, so I'm confused why this is happening. If this works though, then this is fine.

scripts/test_generation.py
mivanit: I think the new interface by @canrager lets us add paths (with formatting) in a more structured way, and then just call `MazePlot().show()`

Note that for testing purposes, tests should call `MazePlot().plot()` and notebooks should call `plt.show()` on a separate line

maze_transformer/generation/generators.py
```diff
@@ -106,7 +106,100 @@ def gen_dfs_with_solution(cls, grid_shape: Coord):
...
+            # Return to the start and retrace our path, connecting cells as we go
+            current = start
+            while not connected[current[0]][current[1]]:
+                direction = direction_matrix[current[0]][current[1]]
+                connected[current[0]][current[1]] = True
+                cells_left -= 1
+
+                next = neighbor(current, direction)
+                # Connect the current and next cell
+                # todo(luciaq) update LatticeMaze to take an adjacency list instead of a connection list for a more
```
mivanit, line 180: I'm not quite sure exactly what this TODO entails, but doesn't `LatticeMaze` have a `from_adj_list()` classmethod?

mivanit (approved):
LGTM! A few questions, but I don't think any changes are needed assuming the merge goes well. I'll confess I haven't fully checked every line of the Wilson's alg implementation but it looks fine.

valedan:
I've just started playing around with this in a notebook, and it seems like the lower right corner is always an island. That doesn't seem right?
Screenshot 2023-04-07 at 11 36 59 AM

valedan:
Correct me if I'm wrong, but from some very quick research I get the impression that Wilson mazes should not have islands at all. I'm seeing lots of islands in these mazes. Unless `MazePlot` is broken somehow? I'm just using `MazePlot(LatticeMazeGenerators.gen_wilson((n, n))).show()`


Screenshot 2023-04-07 at 11 45 32 AM

mivanit:
good catch Dan!!! I was under the impression Wilson's should generate acyclic mazes similarly to DFS. is this incorrect?

valedan:
Yeah I'm pretty sure Wilson mazes should be similar to DFS. See discussion here: https://www.reddit.com/r/proceduralgeneration/comments/g2160z/processing_maze_generation_using_wilsons_algorithm/

luciaquirke:
Yep sorry I just chucked up a fix with a demo notebook, haven't had reception for a bit

luciaquirke:
Screenshot 2023-04-08 at 5 06 49 pm
example

maze_transformer/generation/generators.py
luciaquirke: Yep it does! I was thinking of adding a LatticeMaze initialisation method that takes an adjacency list instead of a connection list because it's the standard data structure used for most maze generation implementations I've seen, but after coding this generation algorithm up it doesn't seem worth it anymore, the data structures are quite similar

scripts/test_generation.py
luciaquirke: I don't have enough context to understand but I assume it's okay for me to go ahead and fix this

luciaquirke: The new interface sounds great. Can should remove any redundant methods and their tests

maze_transformer/generation/latticemaze.py
luciaquirke: Yeah I don't know enough about python config to know why but error went away when I moved it :D

scripts/test_generation.py
mivanit: the `.plot()` method does call the `_plot_maze()`, so I think the tests should stay. surprised this wasnt caught on his PR!

maze_transformer/generation/generators.py
mivanit: are you saying the constructor should be changed to also accept the adjacency list instead of having that only be in a separate method?

Issue #150, opened 2023-03-31
add testing of notebooks in CI


mivanit:
this is being worked on in [add-notebook-testing](https://github.com/AISC-understanding-search/maze-transformer/tree/add-notebook-testing)

# steps:
- [x] [notebook conversion script](https://github.com/AISC-understanding-search/maze-transformer/blob/add-notebook-testing/tests/helpers/convert_ipynb_to_script.py)
- [x] convert and run notebooks from makefile
- [x] disable all plotting in converted notebooks 
  - [x] matplotlib
  - [x] circuitsvis
  - [~] plotly? (this can wait until we actually use plotly somewhere in notebooks)
- [x] add a "default" model to the git repo and/or wandb database (worked on by @luciaquirke )
- [x] add `make test_notebook` to CI
- [x] merge PR :')

# why not `nbconvert`?

(responding to a question by @valedan)

- installing jupyter in CI will add a lot of time -- jupyter does many many things, and even `nbconvert` also includes utilities which we dont need (like converting to PDF)
- actually running `nbconvert` is quite slow! on my machine it was ~2sec per notebook. With just 3 notebooks it might not be a huge deal, but hopefully many notebooks from the experiments repo will eventually migrate here
- the most important reason: even if we use `nbconvert`, we would still have to go into the generated file and modify things to prevent plots from being displayed in CI. The [notebook conversion script](https://github.com/AISC-understanding-search/maze-transformer/blob/add-notebook-testing/tests/helpers/convert_ipynb_to_script.py) would be half as long if it only included processing notebooks, and shorter still if it didn't have code to process all files in a directory (which would just mean it lives in a different file)

rusheb:
FYI there is a tool called nbmake which might do some of this for us.

https://pypi.org/project/nbmake/

mivanit:
> FYI there is a tool called nbmake which might do some of this for us.
> 
> https://pypi.org/project/nbmake/

I don't think there's a way to disable "matplotlib in a widget" for this, so it seems like we'd have to write some kind of script to process it anyway. Thanks for the suggestion though!

PR #153, opened 2023-04-03
Add notebook testing

mivanit:
see #150 

general outline:
- `tests/helpers/convert_ipynb_to_script.py` script:
  - reads notebooks as json, converts to python scripts by stitching cells together
  - (optionally) does cursed things to disable actual plotting for CI
- makefile updates:
  - `help` recipe for printing all recipes
  - `clean` recipe for local testing
  - `convert_notebooks` for running the above script, puts them in `tests/_temp/notebooks`
  - `test_notebooks` for running each of those scripts, and printing output to console if tests fail
  - `test` now defined as `test: clean unit integration test_notebooks`
- various updates to notebooks and related plotting code (mostly as a result of merges)


mivanit:
@luciaquirke @valedan ok this is ready for review for real now I promise

valedan:
After cloning this branch, I get an error when trying to run the `make test_notebooks` command:

```
❯ make test_notebooks
realpath: illegal option -- -
usage: realpath [-q] [path ...]
convert notebooks in notebooks using tests/helpers/convert_ipynb_to_script.py
python tests/helpers/convert_ipynb_to_script.py notebooks/ --output_dir tests/_temp/notebooks --disable_plots
        Converting plot_attention.ipynb...
/Users/dan/active_projects/understanding-search/maze-transformer/tests/helpers/convert_ipynb_to_script.py:120: UserWarning: could not find where circuitsvis is used, plot disabling might not work!
  warnings.warn(
        Converting eval_model.ipynb...
        Converting demo_latticemaze.ipynb...
test notebooks in notebooks
Testing notebooks in tests/_temp/notebooks
  Running tests/_temp/notebooks/demo_latticemaze.py
  Output in tests/_temp/notebooks/demo_latticemaze__CI-output.txt
/bin/sh: /tests/_temp/notebooks/demo_latticemaze__CI-output.txt: No such file or directory
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
Error in tests/_temp/notebooks/demo_latticemaze.py :
cat: tests/_temp/notebooks/demo_latticemaze__CI-output.txt: No such file or directory
make: *** [test_notebooks] Error 1
```

valedan:
I think somewhere in the code may be assuming the presence of a dir?

makefile
```diff
@@ -1,19 +1,84 @@
...
+.PHONY: convert_notebooks
+convert_notebooks:
+	@echo "convert notebooks in $(NOTEBOOKS_DIR) using tests/helpers/convert_ipynb_to_script.py"
+	python tests/helpers/convert_ipynb_to_script.py notebooks/ --output_dir $(CONVERTED_NOTEBOOKS_TEMP_DIR) --disable_plots
+
+.PHONY: test_notebooks
+test_notebooks: convert_notebooks
+	@echo "test notebooks in $(NOTEBOOKS_DIR)"
+	@echo "Testing notebooks in $(CONVERTED_NOTEBOOKS_TEMP_DIR)"
+	@for file in $(CONVERTED_NOTEBOOKS_TEMP_DIR)/*.py; do \
```
valedan, line 48: Not a huge deal, but I wonder if we should put this logic in a python script to simplify the makefile.

maze_transformer/training/config.py
```diff
@@ -149,6 +156,20 @@ class TrainConfig(SerializableDataclass):
         print_loss_interval=50,
         checkpoint_interval=10000,
     ),
+    TrainConfig(
```
valedan, line 159: Is this supposed to be in here? Seems unrelated.

tests/helpers/convert_ipynb_to_script.py
```diff
@@ -0,0 +1,346 @@
...
+            script_lines_new[: cirv_last_import_index + 1]
+            + DISABLE_PLOTS["circuitsvis"]
+            + script_lines_new[cirv_last_import_index + 1 :]
+        )
+        result_str_TEMP = "\n\n".join(script_lines_new)
+
+    return script_lines_new
+
+
+def convert_ipynb(
```
valedan, line 135: I haven't looked at this in detail, but I'm wondering if there's any opportunity to use `nbconvert` to do some of the heavy lifting here? https://nbconvert.readthedocs.io/en/latest/ Always better if we can accomplish the same thing with less new code!

Usage is `jupyter nbconvert --to python notebooks/eval_model.ipynb`

valedan: Ah I just saw this comment discussing the above https://searchingforsearch.slack.com/archives/C04S0FXBF4P/p1680223916440399?thread_ts=1680221802.003669&cid=C04S0FXBF4P

What was the reason to avoid `nbconvert` here? To avoid installing jupyter in CI?

makefile
mivanit: Yeah, that would probably make sense! bash is messy and doubly so when in a makefile. I'll look into this.

tests/helpers/convert_ipynb_to_script.py
```diff
@@ -0,0 +1,346 @@
...
+        - `disable_plots: bool = False`: Disable plots in the output script.
+        - `filter_out_lines: str|typing.Sequence[str] = ('%', '!')`: comment out lines starting with these strings (in code blocks).
+            if a string is passed, it will be split by char and each char will be treated as a separate filter.
+    """
+
+    assert os.path.exists(input_dir), f"Directory {input_dir} does not exist."
+    assert os.path.isdir(input_dir), f"Path {input_dir} is not a directory."
+
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
```
mivanit, line 253: re: https://github.com/AISC-understanding-search/maze-transformer/pull/153#issuecomment-1500459852 @valedan this line should handle it, I think?

mivanit: [c636566](https://github.com/AISC-understanding-search/maze-transformer/pull/153/commits/c636566a49403b52bf3401fbb17a14f6cc98e8a1) sets `exist_ok=True`, but that won't resolve the issue you're having. It works in CI so I'm really not sure what could be the problem

maze_transformer/training/config.py
mivanit: yeah, you're right. that was when I was also planning to include another trained model with multiple checkpoints in the repo, but then decided this should wait until after #157 and forgot about it. I can remove it, but it shouldn't cause any issues for the moment.

tests/helpers/convert_ipynb_to_script.py
mivanit: good question! My reasons are as follows: (also adding to #150 for future reference)

- installing jupyter in CI will add a lot of time -- jupyter does many many things, and even `nbconvert` also includes utilities which we dont need (like converting to PDF)
- actually running `nbconvert` is quite slow! on my machine it was ~2sec per notebook. With just 3 notebooks it might not be a huge deal, but hopefully many notebooks from the experiments repo will eventually migrate here
- the most important reason: even if we use `nbconvert`, we would still have to go into the generated file and modify things to prevent plots from being displayed in CI. The file would be half as long if it only included processing notebooks, and shorter still if it didn't have code to process all files in a directory (which would just mean it lives in a different file)

tests/helpers/convert_ipynb_to_script.py
mivanit: Actually this is probably related to the warning:  
`realpath: illegal option -- -`

at the beginning of the makefile, we define  
`ROOT_RELATIVE_TO_NOTEBOOKS := $(shell realpath --relative-to=$(NOTEBOOKS_DIR) .)`
which is then used to specify output file location in the `test_notebooks` recipe

I assumed that whatever version of `realpath` I have and the CI machine has would be universal, but I guess that's not the case! I think the easiest fix would be to move the majority of `test_notebooks` to a script as per [your suggestion](https://github.com/AISC-understanding-search/maze-transformer/pull/153#discussion_r1160834381), since `ROOT_RELATIVE_TO_NOTEBOOKS` isn't used anywhere else.

mivanit: @valedan , this should be fixed as of [7307f05](https://github.com/AISC-understanding-search/maze-transformer/pull/153/commits/7307f055a34b12256e2738ac7060ad676c365116)

Let me know if this works on your machine!

makefile
```diff
@@ -1,19 +1,80 @@
-format:
```
luciaquirke, line 1: optional question: Are there big differences between using a makefile vs. setuptools for command/build management?

makefile
```diff
@@ -1,19 +1,80 @@
...
-	poetry run python -m pycln --all .
-	poetry run python -m isort format .
-	poetry run python -m black .
+HELPERS_DIR := tests/helpers
+NOTEBOOKS_DIR := notebooks
+CONVERTED_NOTEBOOKS_TEMP_DIR := tests/_temp/notebooks
+POETRY_RUN_PYTHON := poetry run python
```
luciaquirke, line 4: nit/low confidence: don't love this variable because makefiles don't have hover definitions so you have to scroll around for it, and it's as long as the actual command

makefile
```diff
@@ -1,19 +1,80 @@
...
+
 
-test: unit integration
+# listing targets, from stackoverflow
+# https://stackoverflow.com/questions/4219255/how-do-you-get-the-list-of-targets-in-a-makefile
+.PHONY: help
+help:
+	@echo -n "# list make targets"
+	@echo ":"
+	@cat Makefile | sed -n '/^\.PHONY: / h; /\(^\t@*echo\|^\t:\)/ {H; x; /PHONY/ s/.PHONY: \(.*\)\n.*"\(.*\)"/    make \1\t\2/p; d; x}'| sort -k2,2 |expand -t 30
```
luciaquirke, line 80: good thing this is commented 😂 😂

maze_transformer/evaluation/plot_maze.py
```diff
@@ -66,20 +69,26 @@ class StyledPath(PathFormat):
 
 
 def process_path_input(
-    path: CoordArray | StyledPath,
+    path: CoordList | CoordArray | StyledPath,
```
luciaquirke, line 72: Can we unify CoordList and CoordArray? Any reason why we need both?

maze_transformer/evaluation/plot_maze.py
```diff
@@ -218,12 +227,11 @@ def plot(self, dpi: int = 100, title: str = "") -> MazePlot:
         self.ax.set_ylabel("row")
         self.fig.suptitle(title)
 
-        return self
-
     def show(self, dpi: int = 100, title: str = "") -> None:
-        """Plot the maze and paths and show the plot."""
+        """Plot the maze and paths and show the plot. DONT USE THIS IN NOTEBOOKS WHICH NEED TO BE TESTED IN CI!!!"""
```
luciaquirke, line 231: 🤔 Can we display mazes in notebooks which are tested in CI some other way?

maze_transformer/generation/constants.py
```diff
@@ -4,6 +4,7 @@
 Coord = NDArray["x y", np.int8]
 CoordTup = tuple[int, int]
 CoordArray = NDArray["coords", np.int8]
+CoordList = list[CoordTup]
```
luciaquirke, line 7: Can we replace these with tuple[int, int] and list[tuple[int, int]]? If not I think there should be two custom types at most rather than 4

maze_transformer/utils/notebook_utils.py
```diff
@@ -17,16 +23,17 @@ def configure_notebook(seed=42, dark_mode=True):
...
-        ipython.magic("load_ext autoreload")
-        ipython.magic("autoreload 2")
-
-    # Specify plotly renderer for vscode
-    pio.renderers.default = "notebook_connected"
-
-    if dark_mode:
-        pio.templates.default = "plotly_dark"
-        plt.style.use("dark_background")
+    if IN_JUPYTER:
```
luciaquirke, line 26: 🚀

notebooks/eval_model.ipynb
```diff
@@ -33,7 +33,7 @@
     "from maze_transformer.utils.notebook_utils import configure_notebook\n",
     "from maze_transformer.generation.lattice_maze import LatticeMaze, SolvedMaze\n",
     "from maze_transformer.evaluation.plot_maze import PathFormat, MazePlot\n",
-    "from maze_transformer.training.mazedataset import MazeDataset, MazeDatasetConfig\n",
+    "from maze_transformer.training.maze_dataset import MazeDataset, MazeDatasetConfig\n",
```
luciaquirke, line 36: 🙏

luciaquirke (approved):
I haven't reviewed run_notebooks_tests and convert_ipynb_to_script in-depth but think it all looks pretty good and should be easy to modify in the future if necessary since no other component should depend on this code. Thanks for clearing up the notebook imports I broke! Change looks great.

maze_transformer/evaluation/plot_maze.py
mivanit: I should add this to the comment, but in essence `MazePlot` has fig, ax objects and `MazePlot.plot()` will construct the figure. Then, just `pkt.plot()` plots it. Is this what you meant?

makefile
mivanit: I wasnt aware we could use setuptools for this, so I dont really know

makefile
mivanit: They're all short, but I moved them into variables because they are reused multiple times

maze_transformer/evaluation/plot_maze.py
mivanit: Very much agree that these should be unified! Beyond the scope of this PR but tackling it in the next one

maze_transformer/generation/constants.py
mivanit: I'll be refactoring this to unify the types in my next PR, saving that can of worms for later

mivanit:
I forgot to squash im sorry 😭

PR #156, opened 2023-04-06
Baseline model

valedan:
Add a baseline "model". The model inherits from `HookedTransformer` and has a `generate` function which has a compatible interface with `HookedTransformer.generate`, meaning it can be passed into `evaluate_model` just fine. 

The model uses a very simple algorithm to generate predictions: at each step it randomly selects an unvisited neighbouring node until it finds the target or reaches a dead end.

The model accepts a bias parameter to control how likely it is to choose the correct step at each fork.

Demo notebook [here](https://github.com/AISC-understanding-search/experiments/blob/main/dan/00_baseline_evals.ipynb).

Example paths:
Screenshot 2023-03-29 at 11 39 33 AM

## Evaluations
Running `evaluate_model` on a `RandomBaseline` using a 10k 4x4 maze dataset gives the following results:

----node_overlap----
Mean: 0.5930349117549116
Median: 0.5555555555555556

----num_connections_adjacent_lattice----
Mean: 4.2891
Median: 4.0

----fraction_connections_adjacent_lattice----
Mean: 0.7553681349206349
Median: 0.8

----num_connections_adjacent----
Mean: 4.2891
Median: 4.0

----fraction_connections_adjacent----
Mean: 0.7553681349206349
Median: 0.8


mivanit:
RE: failing unit test on [4db3d8c](https://github.com/AISC-understanding-search/maze-transformer/pull/156/commits/4db3d8c38ee87d0c0cc8681715d0c1500ed4e74d)

```
FAILED tests/unit/maze_transformer/evaluation/test_eval_model.py::test_model_loading - RuntimeError: Error(s) in loading state_dict for HookedTransformer:
	Unexpected key(s) in state_dict: "blocks.0.ln1.w", "blocks.0.ln1.b", "blocks.0.ln2.w", "blocks.0.ln2.b", "blocks.1.ln1.w", "blocks.1.ln1.b", "blocks.1.ln2.w", "blocks.1.ln2.b", "ln_final.w", "ln_final.b".
============= 1 failed, 59 passed, 2 skipped, 7 warnings in 24.66s =============
make: *** [makefile:13: unit] Error 1
Error: Process completed with exit code 2.
```

I think I have fixed these issues in the `zanj-integration-2` branch. The old `load_model_with_configs` now has an option for loading with or without folding layernorm, but also this should be resolved completely when using `ZanjHookedTransformer` since it will store whether or not layer norms are folded, and appropriately init the state dict differently.

valedan:
@mivanit Thanks for the explanation! So I should change the baseline to inherit from `ZanjHookedTransformer`?

mivanit:
> @mivanit Thanks for the explanation! So I should change the baseline to inherit from `ZanjHookedTransformer`?

this depends on #157 being merged, so probably not yet. The interface change is also mostly for saving models (not calling them) which isn't really that relevant here

maze_transformer/utils/token_utils.py
```diff
@@ -46,24 +46,40 @@ def get_tokens_up_to_path_start(
...
+def apply_mapping(
+    iter: Iterable[Any],
+    mapping: dict[Any, Any],
+    when_missing: Literal["except", "skip", "include"] = "skip",
+) -> list[Any]:
+    """Given a list and a mapping, apply the mapping to the list"""
+    output = list()
+    for item in iter:
+        if item in mapping:
+            output.append(mapping[item])
```
luciaquirke, line 58: nit: you can `continue` after this statement and remove the `else:` to reduce the nesting a bit

maze_transformer/utils/token_utils.py
```diff
@@ -46,24 +46,40 @@ def get_tokens_up_to_path_start(
...
+                    raise ValueError(f"item {item} is missing from mapping {mapping}")
                 case _:
-                    raise ValueError(f"invalid value for {when_noncoord = }")
+                    raise ValueError(f"invalid value for {when_missing = }")
     return output
+
+
+def tokens_to_coords(
+    tokens: list[str],
+    mazedata_cfg,  # TODO: cannot type this right now because importing MazeDatasetConfig causes a circular import
```
luciaquirke, line 74: mazedata_cfg -> maze_data_cfg

luciaquirke (approved):
Looking good!

valedan:
@mivanit The failing test was actually just due to me being dumb. 🙂

PR #157, opened 2023-04-06
Zanj integration: Models

mivanit:
> Note: this was merged with `add-notebook-testing`, so #153 should be merged before we tackle this (diff should be much cleaner).

added `class ZanjHookedTransformer(ConfiguredModel, HookedTransformer)` which is a muutils `ConfiguredModel`
- this allows easier saving/loading with ZANJ, and stored much config info
- layernorm folding issues hopefully resolved, since the model now stores whether or not the layernorms were folded while saving
- old `load_model_with_configs` should still work as before! please double check changes to that, just in case
- not yet implemented in the main train script, since we are waiting on many changes by @afspies in experiments repo to be folded into this one


mivanit:
ready for review!

maze_transformer/training/config.py
```diff
@@ -210,7 +216,81 @@ def transformer_config(self) -> HookedTransformerConfig:
...
+        state_dict: dict[str, Any],
+        **kwargs,
+    ) -> None:
+        """this is a wrapper around the _load_state_dict function that allows us to do extra things when loading a state dict"""
+
+        recover_exact: bool = kwargs.get("recover_exact", False)
+        fold_ln: bool = kwargs.get("fold_ln", True)
+
+        if self.zanj_model_config.model_cfg.are_layernorms_folded and fold_ln:
+            raise ValueError(
```
valedan, line 264: Should this be an error or just a noop if we try to do it twice?

valedan: Seems like this would trigger any time we try to load a model that was saved with folded layernorms, since `fold_ln` default to True here? And setting it to False would cause the config to be set to False which we don't want?

maze_transformer/training/config.py
```diff
@@ -210,7 +216,81 @@ def transformer_config(self) -> HookedTransformerConfig:
...
+
+        self.zanj_model_config.model_cfg.are_layernorms_folded = fold_ln
+        self.zanj_model_config.model_cfg.are_weights_processed = not recover_exact
+
+        self.load_and_process_state_dict(
+            state_dict,
+            fold_ln=False,
+            center_writing_weights=not recover_exact,
+            center_unembed=not recover_exact,
+            refactor_factored_attn_matrices=not recover_exact,
```
valedan, line 281: Are we sure we always want to set these 3 props to the same value? In the hackathon, `refactor_factored_attn_matrices=True` caused my model to perform terribly, but everything was fine if I turned it off and left the other 2 on. The TransformerLens docs note that it's an experimental method.

tests/unit/maze_transformer/training/zanj/test_zanj_ht_save_load.py
```diff
@@ -0,0 +1,104 @@
...
+
+from maze_transformer.training.config import (
+    BaseGPTConfig,
+    ConfigHolder,
+    MazeDatasetConfig,
+    TrainConfig,
+    ZanjHookedTransformer,
+)
+
+ZANJ_MODEL_CFG: ConfigHolder = ConfigHolder(
```
valedan, line 19: Unrelated to this PR, but I think `SerializableDataclass` has some issue with its types

Screenshot 2023-04-17 at 7 07 13 AM

tests/unit/maze_transformer/training/zanj/test_zanj_ht_save_load.py
```diff
@@ -0,0 +1,104 @@
...
+        name="test_cfg_save-model",
+        act_fn="relu",  # need an actual function here since HookedTransformer will complain otherwise
+        d_model=128,
+        d_head=64,
+        n_layers=4,
+    ),
+)
+
+MODEL: ZanjHookedTransformer = ZanjHookedTransformer(ZANJ_MODEL_CFG)
+MODEL_C: ZanjHookedTransformer = ZANJ_MODEL_CFG.create_model_zanj()
```
valedan, line 32: Nit: I don't know what `Model_C` is - can you improve the variable name here?

tests/unit/maze_transformer/training/zanj/test_zanj_ht_save_load.py
```diff
@@ -0,0 +1,104 @@
...
+                "are_weights_processed": {"self": False, "other": True},
+            }
+        }:
+            pass
+        else:
+            raise e
+
+    # Random input tokens
+    dataset_cfg = model_a.zanj_model_config.dataset_cfg
+    input_sequence = torch.randint(
```
valedan, line 55: A comment would be nice here explaining what `input_sequence` contains. I know it's randomly sampled tokens from the vocab, but it's not obvious at first glance how many. I think 10 in most cases?

maze_transformer/training/config.py
```diff
@@ -210,7 +216,81 @@ def transformer_config(self) -> HookedTransformerConfig:
...
+
+    def _load_state_dict_wrapper(
+        self,
+        state_dict: dict[str, Any],
+        **kwargs,
+    ) -> None:
+        """this is a wrapper around the _load_state_dict function that allows us to do extra things when loading a state dict"""
+
+        recover_exact: bool = kwargs.get("recover_exact", False)
+        fold_ln: bool = kwargs.get("fold_ln", True)
```
valedan, line 261: Why would we want to set either of these kwargs as part of model loading? They're both specified in the model config already right? What's the use case of changing these during the loading process?

valedan (approved):
Looks good, just had a few questions and minor issues.

maze_transformer/training/config.py
mivanit: Yeah, this is a good idea. I am not sure about the motivation behind folding layernorms when saving/loading, maybe @afspies knows? I think making it a noop (with warning?) would be good

maze_transformer/training/config.py
mivanit: Was not aware of this issue, I inherited the code from the old `load_model_with_config` which had all those hardcoded to true. @afspies, what would make the most sense here for the research team?

tests/unit/maze_transformer/training/zanj/test_zanj_ht_save_load.py
mivanit: this might just be due to me not yet having muutils passing mypy, I'll look into it

maze_transformer/training/config.py
afspies: I was under the impression that none of these affect the computation of the model. It seems this is only true for `center_writing_weights` and partially for `center_unembed`. 

`refactor_factored_attn_matrices` seems very involved and it is not clear to me that it is something we should be using.

I would probably have separate flag for `refactor_factored_attn_matrices` and the other two. It might also be reasonable to just have `center_writing_weights` and `center_unembed` true based on an `eval_flag` as it's not the case that they aren't `exact`

mivanit: so tentatively I have as defaults:
- `recover_exact = False`
- `fold_ln = False`
- `refactor_factored_attn_matrices = False`

and `recover_exact` disables `center_writing_weights` and `center_unembed` if set to true

Alex, what do you mean by "it's not the case that they aren't exact"?

maze_transformer/training/config.py
mivanit: I'll just set it to `False` as default for now I think

tests/unit/maze_transformer/training/zanj/test_zanj_ht_save_load.py
```diff
@@ -0,0 +1,104 @@
...
+    model_a: ZanjHookedTransformer, model_b: ZanjHookedTransformer
+):
+    try:
+        assert_model_cfg_equality(model_a, model_b)
+    except ConfigMismatchException as e:
+        if e.diff == {
+            "model_cfg": {"are_weights_processed": {"self": False, "other": True}}
+        } or e.diff == {
+            "model_cfg": {
+                "are_layernorms_folded": {"self": False, "other": True},
```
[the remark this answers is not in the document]
mivanit: this is for testing saving/loading a model with processed weights/folded layernorm, where the state dict keys wont match. since we store this information in the config, we need to account for that part of the config being different

maze_transformer/training/config.py
mivanit: I am not entirely sure what the use case is, but this was a feature in the old `load_model_with_config`. It seemed unwise to remove the ability to fold the layernorm since I'm not sure what exactly it's for

maze_transformer/training/config.py
afspies: It's probably fine to have the defaults as False, although layernorm folding makes a lot of sense as a default practice.

It should not be the case that `center_writing_weights` or `center_unembed` will change the outputs of the model for a given input (though `center_unembed` will change logits before softmax) and so you aren't really changing the computation performed by the model. Perhaps by "exact" we mean "exactly the same way of expressing the parameters as when the model was saved", so it's not really a problem.

mivanit: yeah, `recover_exact` in this context means "make it such that the state dicts are exactly equal". we have two different comparison method in the tests, one which explicitly compares the config and state dict plus another which just compares config + output equality.

PR #158, opened 2023-04-07
Add tests for path evals and a few new evals

valedan:
Changes:
- Unit tests for all evals
- Removed `len(prediction) <= 1` check as I don't think it is correct behaviour (pending discussion with Micheal on this one)
- Added `exact_path_predicted`, `solution_length`, `streak_length_until_incorrect` evals
- Ensured all evals return floats (not sure this is necessary, but seems best to be consistent in the interface)

### DIFF
--- a/maze_transformer/evaluation/path_evals.py
+++ b/maze_transformer/evaluation/path_evals.py
@@ -89,7 +89,9 @@ class PathEvals:
         """fraction of connections in prediction which are are valid paths on the maze"""
 
         num_connections: float = len(prediction) - 1.0
-        return PathEvals.num_connections_adjacent(maze, prediction) / max(num_connections, 1.0)
+        return PathEvals.num_connections_adjacent(maze, prediction) / max(
+            num_connections, 1.0
+        )
 
     @register_method(evals)
     @staticmethod
--- a/maze_transformer/evaluation/path_evals.py
+++ b/maze_transformer/evaluation/path_evals.py
@@ -23,14 +23,18 @@ class PathEvalFunction(Protocol):
 
 def path_as_segments_iter(path: MazePath) -> Iterable[tuple]:
     """
-    Iterate over the segments of a path.
+    Iterate over the segments of a path (ie each consecutive pair).
     """
-    i: int
     step_start: Coord | CoordTup
     step_end: Coord | CoordTup
     for i, step_start in enumerate(path[:-1]):
         step_end = path[i + 1]
         yield step_start, step_end
+
+
+def is_adjacent(node1: Coord, node2: Coord) -> bool:
+    """Check that the nodes are at most 1 step apart (diagonal steps not possible)"""
+    return np.abs(node1 - node2).sum() == 1
 
 
 class PathEvals:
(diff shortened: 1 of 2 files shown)


maze_transformer/evaluation/path_evals.py
```diff
@@ -41,29 +45,20 @@ class PathEvals:
     @register_method(evals)
     @staticmethod
     def node_overlap(solution: MazePath, prediction: MazePath, **_) -> float:
-        """number of shared nodes (any order) / total number of (unique) nodes"""
-        if len(prediction) <= 1:
```
mivanit, line 45: yeah, not sure why this was even here

maze_transformer/evaluation/path_evals.py
```diff
@@ -41,29 +45,20 @@ class PathEvals:
...
 
-        n_shared: int = 0
         solution_set = {tuple(coord) for coord in solution}
         prediction_set = {tuple(coord) for coord in prediction}
 
-        for coord in solution:
-            if tuple(coord) in prediction_set:
-                n_shared += 1
-        return n_shared / len(solution_set)
+        return len(prediction_set & solution_set) / len(solution_set)
```
mivanit, line 53: very clean!

maze_transformer/evaluation/path_evals.py
```diff
@@ -41,29 +45,20 @@ class PathEvals:
...
 
     @register_method(evals)
     @staticmethod
     def num_connections_adjacent_lattice(prediction: MazePath, **_) -> float:
         """number of the connections in prediction which actually connect nodes that are adjacent on the lattice, ignoring if they are adjacent on the maze"""
-        if len(prediction) <= 1:
-            return 0.0
-
-        n_adj: int = 0
+        n_adj = 0.0
```
mivanit, line 59: no type hints :'(

This is definitely a nitpick but I think it helps us with making mypy checking be useful down the line

maze_transformer/evaluation/path_evals.py
```diff
@@ -97,4 +88,40 @@ def fraction_connections_adjacent(
...
 
-        return PathEvals.num_connections_adjacent(maze, prediction) / len(prediction)
+        num_connections = len(prediction) - 1
+        return PathEvals.num_connections_adjacent(maze, prediction) / num_connections
+
+    @register_method(evals)
+    @staticmethod
+    def exact_path_predicted(solution: MazePath, prediction: MazePath, **_) -> float:
+        """Was the maze successfully solved?"""
+        return float(np.all(solution == prediction))
```
mivanit, line 98: I would add a comment here explaining that `==` doesn't actually crash for arrays of different sizes -- I was certainly under the impression that they do

mivanit (approved):
LGTM! Only real comment is that I would maybe add some type hints.

valedan:
Type inference without a hint - hovering `n_adj` shows the correct type:

Screenshot 2023-04-07 at 11 56 03 AM

Does this work on your editor?

mivanit:
> @mivanit Re: type hints - I might be misunderstanding something, but I thought I put type hints everywhere they were needed (ie all function definitions and anywhere that mypy can't infer types).
> 
> For the example you pointed out above - `n_adj: int = 0` I believe the type hint is not necessary here because mypy correctly [infers](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#type-inference) the type without it, and the hint serves no documentation purpose because it's obvious to any reader what is going on.
> 
> Is there a reason I'm missing for having that specific type hint, or are there other places I should have put some?

I always try to add type hints when defining a variable, since this also serves the added benefit of distinguishing between definitions and assignments for any given variable. Even if the actual type itself is obvious, and the code is simple enough to where no documentation is needed, I think it helps to be consistent in "always type hint on var defn" rather than deciding on a case-by-case basis when it is or is not needed.

Honestly, though, this comes from a place of me being neurotic and not a place of logic. Maybe I subconsciously just miss C++ :')

valedan:
Ah okay fair! I think I personally lean a bit more minimalist with this stuff and like to add hints and comments etc only when necessary, but codebase consistency is more important so I'll add them back in. 🙂

maze_transformer/evaluation/path_evals.py
valedan: This is actually a good catch! It does work (numpy broadcasts), but it is deprecated behaviour and emits a warning. Fixing.

PR #160, opened 2023-04-08
Add wilson with solution helper method

luciaquirke:
- Add a helper method that generates a maze using Wilson's algorithm, and a solution to it using A*. The start and ending points of the path are chosen randomly.
- Update the name of the method to get the solution because I found it confusing - open to anything less wordy

I couldn't use this method's outputs with PlotMaze which made me realise we have a bunch of similar but incompatible types that are keeping me from using the solved maze with the PlotMaze.add_true_path. I think we may need to merge CoordTup and Coord, and replace CoordArray with list(CoordTup). Or at least be consistent with how we do it. Not necessary for this PR though.

### DIFF
--- a/maze_transformer/generation/generators.py
+++ b/maze_transformer/generation/generators.py
@@ -100,9 +100,9 @@ class LatticeMazeGenerators:
         )
 
     @classmethod
-    def gen_dfs_with_solution(cls, grid_shape: Coord):
+    def gen_dfs_with_solution(cls, grid_shape: Coord) -> SolvedMaze:
         maze = cls.gen_dfs(grid_shape)
-        solution = np.array(maze.generate_random_path())
+        solution = maze.get_shortest_path_between_random_points()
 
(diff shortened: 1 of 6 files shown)


luciaquirke:
Bump :)

mivanit (commented):
I think it's cleaner to have a single method to take the name of a generator and a grid shape, for when we add more generation algos. I've got an implementation of this in #177 
```python
def get_maze_with_solution(gen_name: str, grid_shape: Coord) -> SolvedMaze:
    maze: LatticeMaze = GENERATORS_MAP[gen_name](grid_shape)
    solution: CoordArray = np.array(maze.generate_random_path())
    return SolvedMaze.from_lattice_maze(lattice_maze=maze, solution=solution)
```
do you think its a good idea to add that to this PR? either #177 or #162 contains code to run through all generators and test them, which I think is also cleaner than testing them individually.

PR #161, opened 2023-04-08
Update ndarray to jaxtyping equivalent

luciaquirke:
Small code health update, also an excuse to learn about jaxtyping


mivanit:
I was already reworking some of the lattice maze related type hints in `add-maze-from-ascii`, I'll get it working on your branch before merging mine

PR #162, opened 2023-04-10
LatticeMaze updates & extensions

mivanit:
Updated and added a lot of things to `LatticeMaze` and classes that inherit from it.

In particular, a way to construct a maze from pixel or ascii representation -- this one's for you @afspies 

Note that due to a blunder on my part, this might depend on #157 being merged first

- `LatticeMaze` now inherits from `SerializableDataclass`
  - this gets us equality checking and `.diff()` checking
  - also prepares us for migrating dataset save/load to ZANJ
- added `TargetedLatticeMaze` which inherits from `LatticeMaze`
  - has `start_pos` and `end_pos` variables
- refactored `SolvedMaze`
  - now inherits from `TargetedLatticeMaze` instead of being a namedtuple
  - still has `maze` property for backwards compatibility, will warn and return self
  - has `start_pos` and `end_pos` fields computed from `solution` (quite jank)

`LatticeMaze`, `TargetedLatticeMaze`, `SolvedMaze` now all have methods:
 - `to_pixels()`: given `(n,n)` maze, create RGB image of `(n*2+1, n*2+1, 3)`. See `PixelColors` in `lattice_maze.py` for color map
 - `from_pixels()`: reverses `to_pixels()`
 - `to_ascii()`: converts to pixels, and then to ascii via `AsciiChars` in `lattice_maze.py` 
 - `from_ascii()`: reverses `to_ascii()`

Notes:
- tests added for most things
- migration of `MazePlot` to use these methods only partially completed, there is some code duplication
- `SolvedMaze.from_pixels()` required a lot of debugging and may have some errors (in computing the solution path order, probably)


notebooks/demo_latticemaze.ipynb
```diff
@@ -40,6 +40,7 @@
                 "import matplotlib.pyplot as plt\n",
                 "\n",
                 "from maze_transformer.generation.generators import LatticeMazeGenerators\n",
+                "from maze_transformer.generation.lattice_maze import LatticeMaze, TargetedLatticeMaze, SolvedMaze\n",
                 "from maze_transformer.evaluation.plot_maze import MazePlot\n",
```
canrager, line 44: There's probably a good reason for this, but what is the benefit of having three different classes over of having one lattice_maze class with optional variables of start_pos, end_pos, solution_path?

notebooks/demo_latticemaze.ipynb
```diff
@@ -72,9 +73,22 @@
             },
             "outputs": [],
             "source": [
-                "maze = LatticeMazeGenerators.gen_dfs(np.array([4,4]))\n",
+                "N: int = 10\n",
                 "\n",
-                "maze"
+                "maze = LatticeMazeGenerators.gen_dfs(np.array([N,N]))\n",
+                "tgt_maze: TargetedLatticeMaze = TargetedLatticeMaze.from_lattice_maze(maze, (0,0), (N-1,N-1))\n",
+                "solved_maze: SolvedMaze = SolvedMaze.from_targeted_lattice_maze(tgt_maze)\n",
```
canrager, line 80: There's probably a good reason for this, but what is the benefit of having three different classes over of having one lattice_maze class with optional variables of start_pos, end_pos, solution_path?

notebooks/demo_latticemaze.ipynb
```diff
@@ -211,17 +225,20 @@
...
             "source": [
                 "MazePlot(maze).plot()\n",
-                "plt.show()"
+                "plt.show()\n",
+                "\n",
+                "plt.imshow(maze.as_pixels())\n",
+                "plt.show()\n",
+                "\n",
+                "print(maze.as_ascii())"
             ]
```
canrager, line 234: High level question: Where do we want the plotting functionality to live? As part of LatticeMaze or and external plotting class (MazePlot)?

As Michael commented, we currently have redundancy for 
- LatticeMaze.to_pixels and MazePlot.plot()
- LatticeMaze.as_ascii and MazePlot.as_ascii

Discussion:
I agree we dont want to have much overhead if one just wants to display a sample maze on the fly (overhead being creating a separate instance of MazePlot class for instance).
However, we discussed earlier, separating plotting functionality to a separate file than LatticeMaze would allow for better testing and more organized/clear code. What are people's thoughts on this?

canrager:
To make redundancy explicitly:
plt.imshow(LatticeMaze.as_pixels()) == MazePlot(unit_length=2).plot()

Screenshot 2023-04-11 at 13 33 39

notebooks/demo_latticemaze.ipynb
mivanit: this is a really going point, and I considered doing it this way! I'm not 100% sure of this path, but my reasoning mostly boils down to:
- type safety: having to constantly check things like `if maze.start_pos is not None` seems like it would add a lot of code and also make it easy to shoot yourself in the foot. static type checkers will play nicer with having a separate class for this
- existing infrastructure: we already had the separation between `LatticeMaze` and `SolvedMaze`, which a lot of code depended on. replacing each usage of that would complicate things, doubly so if we had to perform the above checks constantly

notebooks/demo_latticemaze.ipynb
mivanit: part of my reason for including the `{as/from}_pixels` and `{as/from}_ascii` in the actual underlying `LatticeMaze` and derived classes is that the pixel and ascii representations might be useful not just for human viewing but also for training models and creating datasets. Notably:
- @afspies needs the `from_ascii` or `from_pixels` so that we can create test mazes from his pygame app -- it would be unwieldy to have this in the mazeplot, which then returns a modified `LatticeMaze` or `TargetedLatticeMaze`
- training a model on a rasterized representation of the maze, like in `to_pixels()`, was briefly discussed as something we might want to do
- a different project I am involved in (solving mazes via an architecture derived from RCNNs with some nicer mathematical properties) will probably make use of the rasterized representation

notebooks/demo_latticemaze.ipynb
```diff
@@ -211,17 +225,20 @@
             "outputs": [],
             "source": [
                 "MazePlot(maze).plot()\n",
-                "plt.show()"
+                "plt.show()\n",
```
canrager, line 228: Should we just get rid of the ```MazePlot.show()``` function?

If I understood correctly, we go with
```
Mazeplot.plot()
plt.show()
```

instead of 

```MazePlot.show()``` for better testing within notebooks.

File: notebooks/demo_latticemaze.ipynb (hunk 3 above)
Comment (canrager) [reply notebooks/demo_latticemaze.ipynb:234]: That makes sense. In that case, we keep the redundancy of the functions below?

- LatticeMaze.as_pixels and MazePlot.plot()
- LatticeMaze.as_ascii and MazePlot.as_ascii

Comment (mivanit) [reply notebooks/demo_latticemaze.ipynb:234]: I think we should definitely keep that as part of the interface to `MazePlot`, but perhaps to minimize code duplication have those functions call the underlying functions for `LatticeMaze` if possible.

File: notebooks/demo_latticemaze.ipynb (hunk 4 above)
Comment (mivanit) [reply notebooks/demo_latticemaze.ipynb:228]: I will go ahead and remove `MazePlot.show()` -- if we consume it somewhere in the experiments repo notebooks, the notebook tests should alert us to any issues

**Comment by mivanit** [comment]
@canrager made most of those changes, ready for re-review!

I also significantly refactored `from_pixels()`, `to_pixels()`, `from_ascii()`, and `to_ascii()` -- they are now methods purely of LatticeMaze, and all cases are handled within that!

**Review by canrager — COMMENTED** [review]
Alright, updates on plotting capabilities look good! I'm not confident in reviewing the full PR though.

File: maze_transformer/generation/lattice_maze.py [hunk 5]
```diff
@@ -137,14 +144,22 @@ def get_coord_neighbors(self, c: Coord) -> CoordArray:
             )
         ]
 
-        return np.array(neighbors)
+        output: CoordArray = np.array(neighbors)
+        if len(neighbors) > 0:
+            assert output.shape == (
```
valedan, line 149: What are we guarding against here?

maze_transformer/generation/lattice_maze.py
```diff
@@ -261,6 +315,10 @@ def from_adj_list(
             connection_list=connection_list,
         )
 
+    # ============================================================
+    # TODO: write a to_tokens method?
```
valedan, line 319: Definitely! 😄 It can probably just call `maze_to_tokens`

valedan: Or maybe we just move `maze_to_tokens` into this class as `to_tokens`? Not sure if we have any need for it being seperate.

maze_transformer/generation/lattice_maze.py
```diff
@@ -306,17 +364,426 @@ def from_tokens(cls, tokens: list[str]) -> "LatticeMaze":
...
+        pixel_grid: PixelGrid = np.full(
+            (*pixel_grid_bw.shape, 3), PixelColors.WALL, dtype=np.uint8
+        )
+        pixel_grid[pixel_grid_bw == True] = PixelColors.OPEN
+
+        if self.__class__ == LatticeMaze:
+            return pixel_grid
+
+        # set endpoints for TargetedLatticeMaze
+        if self.__class__ == TargetedLatticeMaze:
```
valedan, line 414: This is a bit sus. Would it be better to have an `as_pixels` on `LatticeMaze` that just returns the basic version, and `as_pixels` on TargettedMaze and SolvedMaze that override it and provide the extra functionality?

valedan: There are a lot of type errors in the rest of this method because you're trying to do stuff that isn't possible on `LatticeMaze`

valedan: (not sure if my solution makes sense - I think it is good to locate the class-specific functionality with the appropriate class, but not 100% sure what the most pythonic version of that looks like)

maze_transformer/generation/lattice_maze.py
```diff
@@ -306,17 +364,426 @@ def from_tokens(cls, tokens: list[str]) -> "LatticeMaze":
...
+        if show_solution:
+            for coord in self.solution:
+                pixel_grid[coord[0] * 2 + 1, coord[1] * 2 + 1] = PixelColors.PATH
+
+            # set pixels between coords
+            for index, coord in enumerate(self.solution[:-1]):
+                next_coord = self.solution[index + 1]
+                # check they are adjacent using norm
+                assert (
+                    np.linalg.norm(np.array(coord) - np.array(next_coord)) == 1
```
valedan, line 434: Seems valuable to to have a method that takes 2 coords and returns a bool indicating if they're adjacent or not. Maybe this would even make sense as an `is_adjacent` method on a `Coord` class (might be out of scope)

maze_transformer/generation/lattice_maze.py
```diff
@@ -306,17 +364,426 @@ def from_tokens(cls, tokens: list[str]) -> "LatticeMaze":
...
+@serializable_dataclass(frozen=True, kw_only=True)
+class TargetedLatticeMaze(LatticeMaze):
+    """A LatticeMaze with a start and end position"""
+
+    # this jank is so that SolvedMaze can inherit from this class without needing arguments for start_pos and end_pos
+    start_pos: Coord
+    end_pos: Coord
+
+    def __post_init__(self) -> None:
+        # make things numpy arrays (very jank to override frozen dataclass)
```
valedan, line 675: If you need to do this jank, maybe it's a sign that this shouldn't be a frozen dataclass?

maze_transformer/generation/lattice_maze.py
```diff
@@ -306,17 +364,426 @@ def from_tokens(cls, tokens: list[str]) -> "LatticeMaze":
...
+        lattice_dim: int = 2,
+    ) -> None:
+        super().__init__(
+            connection_list=connection_list,
+            generation_meta=generation_meta,
+            start_pos=np.array(solution[0]),
+            end_pos=np.array(solution[-1]),
+            lattice_dim=lattice_dim,
+        )
+        self.__dict__["solution"] = solution
```
valedan, line 730: Also here, seems like this just doesn't need to be a frozen dataclass and this line should be `self.solution = solution`

maze_transformer/evaluation/plot_maze.py
```diff
@@ -131,6 +137,23 @@ def __init__(self, maze: LatticeMaze) -> None:
...
+            self.add_true_path(SolvedMaze.from_targeted_lattice_maze(maze).solution)
+
+        if isinstance(maze, SolvedMaze):
+            self.add_true_path(maze.solution)
+
+    @property
+    def solved_maze(self) -> SolvedMaze:
+        if self.true_path is None:
+            raise ValueError(
+                "Cannot return SolvedMaze object without true path. Add true path with add_true_path method."
```
valedan, line 150: Nitpick: Elsewhere in the repo we call this the "solution", not the "true path". Would be good to have consistent naming.

valedan (approved):
Looks good! Some feedback, but nothing super blocking.

maze_transformer/generation/lattice_maze.py
mivanit: neighbors is initialized as a list, since we don't know the length of it to start with. many of these functions should be vectorized and converted to pure numpy, but thats out of scope for this PR. However, the output of `get_coord_neighbors` is expected to be a numpy array

maze_transformer/generation/lattice_maze.py
mivanit: out of scope for this PR, leaving this for later

maze_transformer/generation/lattice_maze.py
mivanit: discussed this during engineering meeting, but for context: I did originally have 3 separate functions, but that was very messy and resulted in a lot of repeated code. Type checking might complain, an `assert` of the type after the `if self.__class__ == WhicheverClass` might help

maze_transformer/generation/lattice_maze.py
```diff
@@ -306,17 +364,426 @@ def from_tokens(cls, tokens: list[str]) -> "LatticeMaze":
...
-        solution = tokens_to_coords(path_tokens, data_cfg)
+        maze: LatticeMaze = LatticeMaze.from_tokens(tokens)
+        path_tokens: list[str] = get_path_tokens(tokens)
+        solution: list[str | tuple[int, int]] = tokens_to_coords(path_tokens, data_cfg)
+
+        assert len(solution) > 0, f"No solution found: {solution = }"
 
-        return cls(maze, cast(list[CoordTup], solution))
+        try:
+            solution_cast: list[CoordTup] = cast(list[CoordTup], solution)
```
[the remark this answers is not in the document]
mivanit: this was to make type checkers happy! probably not needed until mypy gate set up

maze_transformer/evaluation/plot_maze.py
mivanit: Out of scope for now, this comes from an earlier PR. should change this later.

PR #176, opened 2023-04-13
Refactor find_shortest_path and fix minor bug

rusheb:
Some refactoring I did to help me understand  `find_shortest_path`  whilst reading through it.

Mostly renaming. Also updated a few type hints.

There is **one change in behaviour** which I think fixes a minor bug.

Previously, we were initializing `g_score[c_start]` twice, first to zero (correct) and then to the manhattan distance (incorrect).

I think I have fixed this by initializing g_score and f_score for c_start.

Sorry for mixing bug fixes and refactoring together! I will highlight the behaviour changes in a comment.

### DIFF
--- a/maze_transformer/generation/lattice_maze.py
+++ b/maze_transformer/generation/lattice_maze.py
@@ -143,70 +143,56 @@ class LatticeMaze:
         self,
         c_start: CoordTup,
         c_end: CoordTup,
-    ) -> list[Coord]:
+    ) -> list[CoordTup]:
         """find the shortest path between two coordinates, using A*"""
 
-        g_score: dict[
-            CoordTup, float
-        ] = dict()  # cost of cheapest path to node from start currently known
-        f_score: dict[CoordTup, float] = {
-            c_start: 0.0
-        }  # estimated total cost of path thru a node: f_score[c] := g_score[c] + heuristic(c, c_end)
+        # cheapest currently known path to node from start
+        g_score: dict[CoordTup, float] = {c_start: 0.0}
+        # estimated total cost of path through a node: f_score[c] = g_score[c] + heuristic(c, c_end)
+        f_score: dict[CoordTup, float] = {c_start: self.heuristic(c_start, c_end)}
 
-        # init
-        g_score[c_start] = 0.0
-        g_score[c_start] = self.heuristic(c_start, c_end)
+        nodes_evaluated: set[CoordTup] = set()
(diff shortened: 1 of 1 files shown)


maze_transformer/generation/lattice_maze.py
```diff
@@ -143,70 +143,56 @@ def find_shortest_path(
...
-        g_score: dict[
-            CoordTup, float
-        ] = dict()  # cost of cheapest path to node from start currently known
-        f_score: dict[CoordTup, float] = {
-            c_start: 0.0
-        }  # estimated total cost of path thru a node: f_score[c] := g_score[c] + heuristic(c, c_end)
-
-        # init
-        g_score[c_start] = 0.0
-        g_score[c_start] = self.heuristic(c_start, c_end)
```
rusheb, line 158: Here is the part in the old version where I have changed behaviour (see PR description for details)

maze_transformer/generation/lattice_maze.py
```diff
@@ -143,70 +143,56 @@ def find_shortest_path(
...
-            CoordTup, CoordTup
-        ] = (
-            dict()
-        )  # node immediately preceding each node in the path (currently known shortest path)
-
-        while open_vtx:
+        # cheapest currently known path to node from start
+        g_score: dict[CoordTup, float] = {c_start: 0.0}
+        # estimated total cost of path through a node: f_score[c] = g_score[c] + heuristic(c, c_end)
+        f_score: dict[CoordTup, float] = {c_start: self.heuristic(c_start, c_end)}
```
rusheb, line 152: Here is the part in the new version where I have changed behaviour (see PR description for details)

mivanit (approved):
LGTM! I do think that the starting point thing was a bug although I doubt it will change the behavior since we don't use the `f_score` of the starting node ever (since we do not ever pass through it again).

mivanit:
I feel like this might have some merge conflicts with #162 so I will wait to merge that until this is merged

mivanit:
> I feel like this might have some merge conflicts with #162 so I will wait to merge that until this is merged

actually, it looks like there are no conflicts!

rusheb:
@mivanit, should we close this?

mivanit:
> @mivanit, should we close this?

I don't think this actually ended up ever getting merged? I'll check in the maze-dataset repo tommorow and apply the fix there if needed.

PR #177, opened 2023-04-13
Zanj integration: datasets & training

mivanit:
(this is a mega pr, sorry)
# configs

Modifying configs from the command line is now easier!

- `ConfigHolder.get_config_multisource()`
  - takes one of: config object, a file to read the config from, or a list of names to get presets for each of the sub-configs
  - a dotlist-dict to modify any parameters of the config
- `GPTDataset().to_fname()` used to generate filename for saving config (and also to find matching config to load/download). `MazeDatasetConfig` also implements this in a custom way
- `MazeDatasetConfig` now has a `maze_ctor_kwargs` field, for passing keyword arguments to maze generation (see #183)

# maze dataset

You can now get a `MazeDataset` from just a config -- it will load, download, or generate a dataset on the fly. The mess of ways of storing a dataset we had before is now gone -- a `MazeDataset` contains a list of `SolvedMaze`, and it will return one of those when you call `__getitem__`. We also added filters and fixed some parallelization issues!

- `GPTDataset().from_config()` as a new, simplified version of getting a dataset: simply pass a config, and it will attempt to load from local directory, download, or generate. any of these can be disabled, and kwargs (for things like # of cores to use) are passed down.
- canonical representation of the dataset as list of `SolvedMaze`
- `mazes_objs, mazes_tokens, mazes_array` are now cached properties. they will work, but might be slow due to no parallelization
- `MazeDataset.__getitem__()` now returns a `SolvedMaze`
- `create_dataset()` deprecated but should still work. remove this?
- filtering! you can specify filters in the config under the `applied_filters` field, or you can call `dataset.filter_by.your_filter_func(your_arg=your_val)`. Both of these work the same under the hood.
- can specify in `from_config()` whether to run in parallel or not (default is no). this is useful since for small datasets, parallelization has huge overhead. tests are now much faster.
- there may have been some issues to parallelization and using the same fixed seed across all processes. This was fixed in #183 , but in a hacky way

# training

Models now saved as ZANJ objects, and the command line interface is improved.

- `train()` now:
  - saves models as ZANJ
  - returns the trained `ZanjHookedTransformer`
- `train_model()`:
  - now returns `TrainingResult` which contains output path, model, and eventually logging info perhaps?
  - for config, interface inherited from `ConfigHolder.get_config_multisource()` and kwargs are passed as modification dict

# remaining todos:
- [x] add seed to config
- [x] allow filtering a dataset ~~by path length~~ however you want via `dataset.filter_by.some_function(**kwargs)`
  - [x] add `MazeDataset().custom_maze_filter()` which takes a custom function (operating on mazes) as argument -- this makes it easier to add new filters in notebooks etc
  - [x] `@register_wrap_dataset_filter` wraps a function which takes dataset and kwargs, and returns a dataset. we might want to have a different decorator `register_wrap_solved_maze_filter` which just takes a function `(m: SolvedMaze, **kwargs) -> bool` which is then put inside a regular python `filter()` function
- [x] dependencies
  - [x] depends on merging #157 
  - [x] depends on merging #162 
  - [x] publish updated muutils and bump required version (0.3.9)
- [x] tests
  - [ ] tests for passing configs in various formats to `train_model()`
  - [~] *[not needed, return `SolvedMaze`]* check that conversion to array works correctly
  - [~] *[out of scope for this PR]* make tests use common model, as suggested [here](https://github.com/AISC-understanding-search/maze-transformer/pull/48#discussion_r1130843980) in discussion of #48 
  - [x] test filtering of maze datasets
- [~] [out of scope for this PR] implement downloading of maze datasets from wandb
- [x] speed up generation code
  - [x] re-add parallelization in `MazeDataset.generate()`
  - [~] *[out of scope for this PR]* generally speed up / vectorize generation code
  - [~] *[out of scope for this PR]* if possible: JIT-compile with numba?
- [~] *[out of scope for this PR]* integrating with new training code from @afspies in experiments repo
- [x] augmented maze generation
  - [x] merge #184 into this PR
  - [x] add `maze_ctor_kwargs` to `MazeDatasetConfig`, make backwards compatible
  - [x] fix errors?

# questions:
- [x] how should we handle hashes being included in `GPTDataset.to_fname()` ?
    > resolved: no strong opinions, can change this without too much cost. including hash for now.
- [x] should `MazeDataset.__getitem__()` give a `SolvedMaze`, string or tokenized array?
  > resolved: use `SolvedMaze` everywhere


maze_transformer/training/dataset.py
```diff
@@ -41,6 +58,23 @@ class GPTDatasetConfig(SerializableDataclass):
...
+        dict[typing.Literal["name", "kwargs"], str | dict]
+    ] = serializable_field(default_factory=list)
+
+    def __post_init__(self):
+        assert self.seq_len_min <= self.seq_len_max
+        # if seed set to None, then generate a new random seed
+        if self.seed is None:
+            self.seed = torch.random.seed() % 2**31
+
+        if (DEFAULT_SEED != self.seed) and (GLOBAL_SEED != self.seed):
```
valedan, line 72: I think this will always trigger - you're not checking here that GLOBAL_SEED and DEFAULT_SEED are different.

maze_transformer/training/dataset.py
```diff
@@ -115,17 +158,305 @@ def from_sequences(cls, data: list[ATensor[("tokens")]]) -> "IndexedArray":
...
+        do_generate: bool = True,
+        load_local: bool = True,
+        save_local: bool = True,
+        save_formats: set[SaveFormats] = {SaveFormats.OBJECTS, SaveFormats.TOKENS},
+        zanj: ZANJ | None = None,
+        do_download: bool = True,
+        local_base_path: Path = Path("data/maze_dataset"),
+        **kwargs,
+    ) -> None:
+        """base class for gpt datasets
```
valedan, line 188: Nice docs!

maze_transformer/training/dataset.py
```diff
@@ -115,17 +158,305 @@ def from_sequences(cls, data: list[ATensor[("tokens")]]) -> "IndexedArray":
...
+                "no way to load dataset! you said not to load local, not to download, and not to generate"
+            )
+
+        # try loading
+        if load_local:
+            if (local_base_path / fname).exists():
+                output = cls.read(local_base_path / fname, zanj=zanj)
+                did_load_local = True
+
+        if do_download:
```
valedan, line 283: This shouldn't run if we've already loaded locally.

maze_transformer/training/dataset.py
```diff
@@ -115,17 +158,305 @@ def from_sequences(cls, data: list[ATensor[("tokens")]]) -> "IndexedArray":
...
+            try:
+                output = cls.download(cfg, **kwargs)
+            except NotImplementedError:
+                pass
+
+        if do_generate:
```
valedan, line 289: Shouldn't generate if we've already downloaded.

maze_transformer/training/dataset.py
```diff
@@ -115,17 +158,305 @@ def from_sequences(cls, data: list[ATensor[("tokens")]]) -> "IndexedArray":
...
+
+    sets:
+    ```python
+    dataset_cls._FILTER_NAMESPACE = filter_namespace_cls
+    filter_namespace_cls._BASE_DATASET = dataset_cls
+    ```
+    and also adds a `_filter_namespace_cls` attribute to every static method in the namespace class for use in `register_wrap_dataset_filter`
+    """
+
+    def decorator(filter_namespace_cls: type) -> type:
```
valedan, line 415: Do types need updating here?

maze_transformer/training/dataset.py
```diff
@@ -115,17 +158,305 @@ def from_sequences(cls, data: list[ATensor[("tokens")]]) -> "IndexedArray":
         return cls(arr=arr, indices=indices)
 
 
+class SaveFormats(enum.Enum):
+    OBJECTS: str = "objects"
+    TOKENS: str = "tokens"
+    ARRAY: str = "array"
+
+
 class GPTDataset(Dataset):
```
valedan, line 167: Lots of great stuff in here, but as far as I can tell this class has no unit tests at all? I'd definitely like to see those added before merging.

maze_transformer/training/maze_dataset.py
```diff
@@ -114,265 +125,272 @@ def n_tokens(self) -> int:
...
+        return len(self.mazes)
 
     def get_all_lengths(self) -> list[int]:
-        return self.mazes_array.get_all_lengths().tolist()
+        raise NotImplementedError()
+        # return [len(m) for m in self.mazes_tokens]
 
     @classmethod
-    def gen_default(
+    def generate(
```
valedan, line 215: We need some tests for generation too.

maze_transformer/generation/generators.py
```diff
@@ -201,9 +191,23 @@ def neighbor(current: Coord, direction: int) -> Coord:
...
+
 
 # TODO: use the thing @valedan wrote for the evals function to make this automatic?
 GENERATORS_MAP: dict[str, Callable[[Coord, Any], "LatticeMaze"]] = {
     "gen_dfs": LatticeMazeGenerators.gen_dfs,
     "gen_wilson": LatticeMazeGenerators.gen_wilson,
 }
+
+
+def get_maze_with_solution(gen_name: str, grid_shape: Coord) -> SolvedMaze:
```
mivanit, line 210: @luciaquirke in reference to #160

maze_transformer/training/maze_dataset.py
```diff
@@ -114,265 +125,272 @@ def n_tokens(self) -> int:
...
-                )
+    def __getitem__(self, index: int, pad: bool = True) -> str:
+        """index into mazes_array.arr, getting from the start of the correct sequence, padding if necessary"""
 
-        if self.mazes_objs is not None and self.mazes_array.indices is not None:
-            if len(self.mazes_objs) != len(self.mazes_array.indices):
-                raise ValueError(
-                    f"MazeDataset invalid: {len(self.mazes_objs) = }, {len(self.mazes_array.indices) = }"
-                )
+        return " ".join(maze_to_tokens(self.mazes[index], self.cfg.node_token_map))
```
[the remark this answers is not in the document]
afspies: This is fine, but it means we will need to provide a collation function for the pytorch dataloader to do the batching (as we aren't padding inside of the dataset __getiem__. This is _probably_ the most efficient way to do it tbf. 

That being said, it is fairly amusing that we start from a file of maze strings which we convert to `SolvedMaze` objects, then reconvert those into tokens with `maze_to_tokens` which in turn gets joined into a full string, then collated, and then batched by the `wrapped_tokenizer`.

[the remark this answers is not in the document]
valedan: After playing around with this a bit I'm leaning towards returning a SolvedMaze here, since that's the most generally useful thing - we can have lots of methods on it to easily convert to other formats. Then as you say, do the conversion to tokens in the training loop - this can be more efficient because we can operate on the whole batch.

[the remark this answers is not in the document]
afspies: Indeed It would be good to have a standard representation for mazes which we pass around, and then all the conversion happens inside plotting functions or the tokenizer.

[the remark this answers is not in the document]
valedan: Yep good point - I need to think about how to handle that in a nice way.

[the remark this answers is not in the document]
mivanit: `SolvedMaze` now inherits from `LatticeMaze`, but is assumed to actually have a `solution` component. Maybe then we should return a `LatticeMaze`, and perform a check downstream to ensure the object is actually an instance of `SolvedMaze`?

Question for @afspies : do we want to assume that the mazes come with a start and end point? If so, its trivial to then compute a solution (assuming one is possible).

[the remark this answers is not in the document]
valedan: Hmm yeah if we need to support mazes without solutions then it may indeed be better to roll the `TargettedMaze` and `SolvedMaze` back into `LatticeMaze`. Having them seperate has been causing us headaches anyway - needing to do `if maze.solution` in a bunch of places may be the lesser evil here.

Btw @mivanit I’ve been working on some of these discussed changes [here](https://github.com/AISC-understanding-search/maze-transformer/tree/zanj-datasets-getitem) - haven’t wanted to put the commits in this PR yet because they’re still very WIP, just letting you know so we don’t duplicate effort.

mivanit:
@valedan here are the remaining problems which we need to fix before merging. Once tests pass, I think we are good to go!

- [ ] some kind of transposing issue in baseline solver tests: `tests/unit/maze_transformer/evaluation/test_baseline_models.py`
  - unclear why this is only causing issues with baseline solver
- [ ] `fixture "mocker" not found` in `tests/unit/maze_transformer/training/test_dataset.py`
  - no idea how pytest fixtures work, or what I did in #183 to break this

valedan (approved):
🚀

PR #182, opened 2023-04-24
Zanj datasets getitem

valedan:
Main changes:
- `MazeDataset.__getitem__` returns a `SolvedMaze`
  - Tokenization for the training loop is handled in the dataloader
- Move `maze_to_tokens` to `LatticeMaze.as_tokens()`
- Added `MazeDataset.as_tokens` replacing `mazes_tokens`
- Remove `mazes_objs`, `mazes_tokens`, `mazes_array`
- Add some tests for dataloader, finish tests for MazeDataset
- Ran `make format`


maze_transformer/evaluation/eval_model.py
```diff
@@ -150,15 +149,12 @@ def evaluate_model(
...
     }
 
-    for batch in chunks(dataset.mazes_tokens, batch_size):
-        # TODO: This won't be needed after #124, then we can call mazes_objs instead
-        # https://github.com/orgs/AISC-understanding-search/projects/1/views/1?pane=issue&itemId=23879308
-        solved_mazes: SolvedMaze = [
-            SolvedMaze.from_tokens(tokens, dataset.cfg) for tokens in batch
+    for maze_batch in chunks(dataset, batch_size):
+        tokens_batch = [
+            maze.as_tokens(dataset.cfg.node_token_map) for maze in maze_batch
```
mivanit, line 154: we spoke today about having the tokenization be done by the model (since that is where we have access to the `ConfigHolder`, but changing that can wait

maze_transformer/training/dataset.py
```diff
@@ -112,6 +112,7 @@ def to_fname(self) -> str:
         return sanitize_fname(f"f{self.name}_{self_json_hash}")
 
 
+# TODO: This class is not referenced anywhere. Can we delete?
```
mivanit, line 115: yeah, this class should be deleted. I might add it to muutils though since it seems generally useful.

maze_transformer/training/maze_dataset.py
```diff
@@ -164,52 +162,34 @@ def __init__(
     def data_hash(self) -> int:
         return hash(tuple(self.mazes))
 
+    # TODO: Is this necessary? Objects is handled by getitem, tokens is trivial with maze.as_tokens, array seems to not be needed.
+    # Maybe we do need something to easily get a tokenized single maze without needing to pass in a node token map
```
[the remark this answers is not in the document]
mivanit: neither required for this PR, certainly not migration of tokenization logic

maze_transformer/training/maze_dataset.py
```diff
@@ -381,16 +361,15 @@ def start_end_distance(maze: SolvedMaze, min_distance: int) -> bool:
     @register_wrap_dataset_filter
     @staticmethod
     def cut_percentile_shortest(
-        dataset: MazeDataset, percentile: float = 0.1
+        dataset: MazeDataset, percentile: float = 10.0
```
mivanit, line 364: I absolutely did not realize that `np.percentile` took a percent between 0 and 100. maybe add a comment to the docstring about this, seems like a gotcha?

maze_transformer/training/training.py
```diff
@@ -38,19 +40,23 @@ class TRAIN_SAVE_FILES:
     )
 
 
+def collate_batch(batch: list[SolvedMaze], config: MazeDatasetConfig) -> list[str]:
+    # Perf could be improved by vectorizing this
```
mivanit, line 44: im not sure if making this a list comprehension (I assume this is what you mean by vectorizing given that I dont think this is numpy-vectorizable) would make this faster, but it would be cleaner

maze_transformer/training/training.py
```diff
@@ -88,13 +94,12 @@ def train(
...
     )
     for iteration, batch in enumerate(dataloader):
-        # compute loss
-        # batch_on_device: Int[torch.Tensor, "batch sequence"] = batch.type(
-        #     dtype=torch.LongTensor
-        # ).to(model.cfg.device)
-
-        # loss: Loss = model(batch_on_device[:, :-1], return_type="loss")
-        loss: Loss = model(batch, return_type="loss")
+        loss: Float[torch.Tensor, ""]
```
mivanit, line 97: I don't thinks `loss` is actually a plain tensor here, I'm pretty sure its a `TransformerLens` Loss object?

tests/unit/maze_transformer/generation/test_maze_dataset.py
```diff
@@ -21,45 +27,31 @@ def test_generate_serial(self):
...
+        # TODO: This comparison does not work due to shuffling
+        # assert dataset.get(2, SaveFormats.TOKENS) == dataset[2].as_tokens(dataset.cfg.node_token_map)
 
         with pytest.raises(NotImplementedError):
             dataset.get(2, SaveFormats.ARRAY)
 
         with pytest.raises(ValueError):
-            dataset.get(2, "foo") # type: ignore
-
+            dataset.get(2, "foo")  # type: ignore
```
mivanit, line 54: replace this and other instances if we remove `get()`

mivanit (approved):
LGTM! only minor notes, which can wait until we merge back into `zanj-integration-datasets`

maze_transformer/training/training.py
valedan: I did mean the numpy/pytorch version, and I realize that it would be a big chunk of work... but it seems maybe possible? The connection_list and solution are already numpy arrays, so it seems possible that would could take a batch of maze objects and turn it into a batch of tokens (the integer kind) using only tensor operations. I could totally be wrong about the feasibility here. This would be TONS of work, and my benchmarking shows that this collation contributes less than 10% to the runtime of the overall training loop, so definitely not worth it right now!

maze_transformer/training/dataset.py
```diff
@@ -322,30 +323,13 @@ def _apply_filters_from_config(self):
...
-def register_filter_namespace_for_dataset(dataset_cls: type[GPTDataset]) -> type:
-    """register the namespace class with the given dataset class
+def register_filter_namespace_for_dataset(dataset_cls: Type[GPTDataset]) -> Callable[[Type], Type]:
+    """register the namespace class with the given dataset class """
 
-    sets:
-    ```python
-    dataset_cls._FILTER_NAMESPACE = filter_namespace_cls
-    filter_namespace_cls._BASE_DATASET = dataset_cls
-    ```
```
mivanit, line 411: what's the reason for removing this comment?

valedan: The first part:

```    
```python
    dataset_cls._FILTER_NAMESPACE = filter_namespace_cls
    filter_namespace_cls._BASE_DATASET = dataset_cls
```
```

Is just an exact duplicate of the code a couple lines down, it doesn't explain anything.

The second part

`and also adds a `_filter_namespace_cls` attribute to every static method in the namespace class for use in `register_wrap_dataset_filter``

Is no longer true because I removed that part of the method (as suggested in the TODO comment).

File: maze_transformer/training/training.py (hunk 6 above)
Comment (valedan) [reply maze_transformer/training/training.py:97]: Good point! I think we can narrow it down though, in this case we know it is actually a `SingleLoss` because we're not setting `loss_per_token=True` in our forward call. 

These are the relevant TL types:
```
SingleLoss = Float[torch.Tensor, ""] # Type alias for a single element tensor
LossPerToken = Float[torch.Tensor, "batch pos-1"]
Loss = Union[SingleLoss, LossPerToken]
```

Definitely better to use `SingleLoss` here rather than what I've got, I'll make that change. Also open to changing it back to `Loss` if you think that's best.

PR #185, opened 2023-05-13
Collected Datasets, various fixes

mivanit:
Changes in this PR:

# dataset related

- **added dataset de-duplication filters** (fast and slow, fast sacrifices granularity)
- **added `CollectedMazeDataset` and associated config**
  - config contains a list of `MazeDatasetConfig`
  - samples from the component datasets in order
  - filters applied to datasets included
- **better handling of generation metadata**
  - metadata now properly inherited and saved in `SolvedMaze` and `LatticeMaze`
  - filter to collect (in `collections.Counter` objects) generation metadata in a dataset from the component mazes
  - filter to strip metadata from component mazes (since it takes up a lot of space and we might not need it)
- removed some old dataset methods
- some fixes to ZANJ which were causing issues
- a lot of various tokenization issues and cleanup

# other

- **updates to training code**
  - moved out of scripts and into `training/train_model.py`
  - more logging and options
  - misc cleanup
- **updated the notebooks and related testing**
  - made sure that they were *actually being run during tests*
  - moved some functionality out into repo (particularly `ProcessedMazeAttention`)
- **added example models and dataset to repo**
- updated the readme, added some nice pictures

> this PR was originally motivated by [this slack discussion](https://searchingforsearch.slack.com/archives/C04T4Q89T4Y/p1683914223877189), enabling @afspies to use the dataset for his training sweeps. it ended up being a lot bigger than just that, oops.


maze_transformer/dataset/collected_dataset.py
```diff
@@ -0,0 +1,113 @@
...
+                dataset.mazes for dataset in self.maze_datasets
+            )
+        )
+
+    def __len__(self) -> int:
+        return sum(len(dataset) for dataset in self.maze_datasets)
+
+    def __getitem__(self, index: int):
+        # find which dataset the index belongs to
+        dataset_idx: int = np.searchsorted(self.dataset_cum_lengths, index)
```
valedan, line 71: I'm concerned about the performance here. `searchsorted` uses binary search which is logN, but generally a  `__getitem__` method would be constant time. Since this is used so heavily by the dataloader I'm worried this will tank training performance for non-trivial dataset sizes.

mivanit: I agree that a cleaner solution here would be to concatenate ahead of time, this was must a quick proof of concept.

However, its worth noting that this getitem is actually log(m) where m is the number of datasets, so if m > 100 then probably something very bad is happening

valedan: Oh derp, in that case I'm not concerned at all!

valedan:
@mivanit I've added the first few tests for the collection, more to come. The getitem one is failing, I don't know why.
 
Screenshot 2023-05-14 at 11 58 43 AM

Screenshot 2023-05-14 at 11 56 58 AM

maze_transformer/dataset/maze_dataset.py
```diff
@@ -395,4 +411,173 @@ def cut_percentile_shortest(
...
+        - if two solutions are of different lengths, they will never be considered duplicates
+            TODO: check for overlap?
+        """
+        unique_mazes: list[SolvedMaze] = list()
+
+        maze_a: SolvedMaze
+        maze_b: SolvedMaze
+        for i, maze_a in enumerate(dataset.mazes):
+            a_unique: bool = True
+            for maze_b in dataset.mazes[i + 1 :]:
```
valedan, line 452: This is quadratic in the number of mazes, right? I did some quick benchmarking:
 
100 mazes - 40ms
1000 mazes - 3.4s
10000 mazes - 300s

Looks like this will be too slow for our standard 5 million maze dataset. Since we're trying to get this out quickly maybe we should simplify to just consider exact matches for now so we can use a set?

mivanit: Thanks for benchmarking this! I agree this is way too slow. Let's add a `remove_duplicates_fast` filter which does this?

valedan: Okay I've added this. Dedupes a 100k dataset in 15s and scales ~linearly. But I'm not sure if it is useful. On my 100k dataset there were no exact duplicates for 6x6. I did verify that there are lots of duplicates for 2x2.

PR #188, opened 2023-06-07
Add optional dataset-only dependencies

mivanit:
It would be nice to have a way to install the dataset only, without transformers or TransformerLens packages.


mivanit:
Closing -- #189 exports all dataset functionality to external library, so this is no longer needed.

PR #203, opened 2024-01-26
Add conda instructions to README.md

naveenarun:
I ran into a poetry version issue when working with Conda, so I added that here with `poetry self update` to reduce hiccups.


mivanit (changes requested):
LGTM, but we should resolve the issues with wandb failing in integration tests before merging.

PR #204, opened 2024-01-27
Allow training notebook to run without logger or wandb config

naveenarun:
Added some changes that let the training notebook to run without wandb API; however, I haven't been able to test whether logging works *with* the wandb API, since I don't have access.

Added a unit test as well in `tests/units/maze_transformer/test_offline_training.py`, which checks that you can successfully run `train_model` when `wandb_project` is `None`.


mivanit (approved):
LGTM, will merge to integration branch

PR #208, opened 2024-02-08
Try and fix tokenizer to work with new version of transformers library

afspies:
The proposed solution is probably not backwards compatible, and is fairly hacky (it strips spaces, and I am not sure it properly assigns vocab / special tokens). I suggest engineering handle this properly before merging:

There is an issue with our tokenization in the new version of transformers. In particular, in the tokenize function from transformers.tokenization_utils.py  the line tokens = self.tokens_trie.split(text) returns a list of tokens with spaces if the input sequence is <path_start> (1,0)… (i.e. includes spaces). this wasn’t the case before, and I suspect stems from how I have to change the addition of the vocabulary in our tokenizer (to work with their new way of handling token addition via the _add_tokens method (we can’t just overwrite the dicts as these are now properties >.<). As a temporary fix we can manually remove spaces from sequences, but that’s quite disgusting

The best option might be to create token jsons and push a tokenizer to huggingface.


aaron-sandoval:
I branched [tokenizer-fix-aaron](https://github.com/understanding-search/maze-transformer/tree/tokenizer-fix-aaron) from [tokenizer-fix](https://github.com/understanding-search/maze-transformer/tree/tokenizer-fix) to work on this issue. Before looking at the tokenization behavior you discuss above, I'm starting with getting `transformers` to stop looking for our tokenizer from the HF hub.

aaron-sandoval:
Notes from diving down stack traces:

- Calling `maze_transformer.training.config.ConfigHolder.tokenizer` never assigns the result of the constructors to `self._tokenizer`. Repeated calls build new tokenizers every time.
  - New tokenizer was built 5 times during 1 call to `ZanjHookedTransformer.read`
- Bug: trying to download tokenizer from HF
  - Added a kwarg `self.init_kwargs["add_bos_token"] = True` in `HuggingMazeTokenizer.__init__`. 
  - `eval_model.ipynb` runs and outputs look fine
  - Unit tests all run and 3 fail
    - No more errors with trying to download tokenizer from HF
    - `test_tokenizers.test_manual_tokenizer` failed:Looks like it's because `MazeTokenizer` in the MazeDataset library didn't implement `get_vocab`, even though `HuggingMazeTokenizer` did.
- Haven't gotten anywhere near the Trie behavior Alex was talking about. But `eval_model.ipynb` running and lots of unit tests passing seems encouraging in my ignorance.

aaron-sandoval:
More notes:
- All 3 unit test failures trace to calls to `ConfigHolder.tokenizer`
  - When it enters the conditional on line 476:
    - `return PreTrainedTokenizer(**self.pretrainedtokenizer_kwargs)`
  - Resolving this one conditional might be enough to pass all unit tests
- `transformers.PreTrainedTokenizer` cannot now be directly instantiated via `__init__` without subclassing
  - `__init__` unconditionally calls `get_vocab` which raises `NotImplementedError`
  - Any direct instantiations of `PreTrainedTokenizer` without subclassing will fail
    - `ConfigHolder.tokenizer` is only instance I found
      - Can the conditional which calls `PreTrainedTokenizer` be changed?
      - Or delete that if clause and give `ConfigHolder` a default `maze_tokenizer`?
      - What is the tokenizer produced by this conditional clause supposed to do if there's no vocab reference?
- Why does `ConfigHolder.tokenizer` not cache the tokenizers it builds on every call?
  - The only reason I can think of is if the tokenization behavior needs to dynamically respond to changes in mutable `ConfigHolder` properties. Is this right?

aaron-sandoval:
Notes on new tokenizer behavior:
- `test_random_baseline` unit tests still failing due to tokenizer behavior, not calls to `ConfigHolder.tokenizer`
- Changes in tokenizer behavior
  - First differences emerge in `PreTrainedTokenizer.tokenize`
    - No difference in Trie outputs (for one input in `test_random_baseline` that I tracked in the debugger
    - Small change in how the Trie outputs (`list[str]`) are stripped internally in `PreTrainedTokenizer.tokenize`
    - v4.33 `PreTrainedTokenizer.tokenize`:
      - Returns list of split strings with spaces from the `text` input stripped to empty strings
        - Ex: `['<ADJLIST_START>', '', '(0,0)', '', '<-->', ...]`
      - Empty strings are skipped over
    - v4.38 `PreTrainedTokenizer.tokenize`:
      - Returns list of split strings with spaces from the `text` input converted to **2** empty strings each
        - Ex: `['<ADJLIST_START>', '', '', '(0,0)', '', '', '<-->', ...]`
      - With the internal change in string stripping, it calls `HuggingMazeTokenizer._tokenize(' ')`, which currently returns `['', '']`
        - I'll try changing `HuggingMazeTokenizer._tokenize(' ')` to emulate the old output

tests/unit/maze_transformer/test_tokenizers.py
```diff
@@ -117,6 +119,80 @@ def test_to_ascii(tok_mode):
...
+                ],
+            ],
+            [
+                TokenizationMode.AOTP_UT_uniform,
+                TokenizationMode.AOTP_UT_rasterized,
+            ],
+        )
+    ],
+)
+def test_maze_to_tokens_roundtrip(maze_ascii: list[str], tok_mode: TokenizationMode):
```
mivanit, line 172: only thing I can think of is that it would be good to test that tokenization of these mazes equals some hand-inspected string

maze_transformer/tokenizer.py
```diff
@@ -113,12 +112,20 @@ def __call__(self, text, **kwargs) -> BatchEncoding:
...
 
         return text.split(" ")
 
     def _convert_token_to_id(self, token: str) -> int:
         if token in self.vocab:
             return self.vocab[token]
+        elif (
+            token == " "
+        ):  # for some reason transformers trie now returns ' ' as tokens
+            return None
```
mivanit, line 128: this feels hacky -- does this case ever come up in practice?

mivanit (approved):

mivanit:
> ### Caching tokenizers in `ConfigHolder.tokenizer`
> * I tried caching the `HuggingMazeTokenizer` built on line 478 in `self._tokenizer`
> * This caused a few other unit tests to fail, so I reverted it

No idea why this is happening -- what exactly failed?

> ### Changes to tests
> * Test description per @mivanit
>   * get a maze from an ascii representation
>   * tokenize it
>   * create a maze from the tokenized representation, check that they match
>   * check that all non-adjacency list tokens match

The current implementation looks good, but it would be good to compare to a hand-inspected example (i.e. to make sure we aren't placing multiple special tokens)

> * `test_maze_to_tokens_roundtrip` implementation
>   * Unclear which methods should be used
>   * Easiest implementation of the above uses all `maze_dataset` code and none of the code updated in this PR
>   * @mivanit What was the intent of this test?

Now that you mention this only uses `maze-dataset` code, maybe this is a test it makes sense to add to that repo instead.

> * Existing `test_tokenizer_inside_hooked_transformer` has test coverage of modified code
>   * `PreTrainedTokenizer.tokenize`
>   * `HuggingMazeTokenizer._tokenize`
>   * `HuggingMazeTokenizer.__call__`
>   * `transformer_lens.utils.get_tokenizer_with_bos`
> * @mivanit Anything else is needed in tests?

I think this looks sufficient!

> ### @afspies' Changes
> * Reverted his changes to `HuggingMazeTokenizer.__call__` since it's no longer necessary
> * Haven't looked into his other changes in [35be002](https://github.com/understanding-search/maze-transformer/pull/208/commits/35be002db6f1d6d7c9f51995dcf127537901333b)
> * @mivanit If the tests pass, should we still look into those other changes?

If they pass, ship it 💯 (I'm pretty confident about our notebook tests catching if something horribly breaks_

PR #209, opened 2024-02-23
Fix tokenizer to allow update to `transformers >=4.34.0`

aaron-sandoval:
This is a sub-branch for now to not mess with Alex's [tokenizer-fix](https://github.com/understanding-search/maze-transformer/tree/tokenizer-fix) branch.


aaron-sandoval:
Notes from diving down stack traces:

- Calling `maze_transformer.training.config.ConfigHolder.tokenizer` never assigns the result of the constructors to `self._tokenizer`. Repeated calls build new tokenizers every time.
  - New tokenizer was built 5 times during 1 call to `ZanjHookedTransformer.read`
- Bug: trying to download tokenizer from HF
  - Added a kwarg `self.init_kwargs["add_bos_token"] = True` in `HuggingMazeTokenizer.__init__`. 
  - `eval_model.ipynb` runs and outputs look fine
  - Unit tests all run and 3 fail
    - No more errors with trying to download tokenizer from HF
    - `test_tokenizers.test_manual_tokenizer` failed:Looks like it's because `MazeTokenizer` in the MazeDataset library didn't implement `get_vocab`, even though `HuggingMazeTokenizer` did.
- Haven't gotten anywhere near the Trie behavior Alex was talking about. But `eval_model.ipynb` running and lots of unit tests passing seems encouraging in my ignorance.

PR #214, opened 2024-05-17
Support maze dataset PR #37: Tokenizers Step 2

aaron-sandoval:
Updatee to support [maze-dataset PR #37](https://github.com/understanding-search/maze-dataset/pull/37) when merged.

### DIFF
--- a/maze_transformer/evaluation/baseline_models.py
+++ b/maze_transformer/evaluation/baseline_models.py
@@ -11,7 +11,7 @@ from maze_dataset import (
     LatticeMaze,
     SolvedMaze,
 )
-from maze_dataset.tokenization.token_utils import (
+from maze_dataset.token_utils import (
     get_origin_tokens,
     get_path_tokens,
     get_target_tokens,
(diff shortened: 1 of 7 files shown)


maze_transformer/training/train_save_files.py
```diff
@@ -6,7 +6,7 @@
 from maze_transformer.training.config import ConfigHolder
 
 
-@freeze
```
aaron-sandoval, line 9: It looks like `freeze` can only act on objects, not types. It raises an exception when it recursively calls `freeze(TRAIN_SAVE_FILES.__dict__)`, since it can't act on a `mappingproxy`.

aaron-sandoval:
@mivanit All but 2 unit tests are fixed. The last 2 are both zanj tests comparing model outputs before/after a round trip save/load. I'm a bit at a loss on these, I think they could use your eyes. Since 2/4 of the zanj tests pass, it seems more likely due to some other update rather than anything to do with tokenizers.

mivanit:
so integration tests [are hanging](https://github.com/understanding-search/maze-transformer/actions/runs/10478671915/job/29022569828) after they are complete. 

pretty sure I've narrowed this down to a wandb issue, since [8311eb8](https://github.com/understanding-search/maze-transformer/pull/214/commits/8311eb8ae3c35b200c00935185ed5edfb60770ac) makes `test_eval_model.py` not hang just by moving `test_model_loading`

mivanit (approved):

PR #215, opened 2024-06-06
allow HookedTransformer without ZanjHookedTransformer

junyuanjyjy:
Tokenizer fix similar to commit on 'eval_single_token_tasks.py'


mivanit (commented):
good catch!

mivanit (approved):
LGTM, good catch!

